PR #6, opened 2016-04-10
Added missing files from SDK

AdySan:
Wire library examples wouldn’t compile without these. Specifically `nrf_twis.h` includes `nrf_drv_config.h` which is missing from the core. Probably need a custom `nrf_drv_config.h` for nRF52DK though.


sandeepmistry:
@AdySan thank you for submitting this.

However, I don't believe this file is needed, so I've stubbed it for now in https://github.com/sandeepmistry/arduino-nRF5/commit/f476e0ce355fbc223f354882a2e9ec074ee74a92.

PR #19, opened 2016-07-05
Add platform.txt recipe for 'Export compiled binary' as a hex file

rogerclarkmelbourne:
Sandeep, I added the 2 lines needed for the IDE to support the "Export compiled Binary" menu feature, (on the Sketch menu).

I've set this to use the hex version of file rather than the binary (bin) version, as the Hex file has the start offset in it, which will be needed if anyone wants to upload using JFlash etc


rogerclarkmelbourne:
I've now done a second PR as I've just added a new variant for the Waveshare BLE400 nRF51822 evaluation kit
http://www.waveshare.com/product/modules/wireless/bluetooth/nrf51822-eval-kit.htm
LED's and Buttons have been defined in variant.h and have been tested.
SPI and I2C have also been defined, though I've only tested these as GPIO using a multimeter, but the pin numbers seem to be correct.
Serial pins have been mapped to its onboard USB to Serial and I have tested both send and receive functions

PR #26, opened 2016-07-28
support x86_64 Linux

morganrallen:
Prevents error "Tool openocd is not available for your operating system." on 64 bit Linux systems.


sandeepmistry:
@morganrallen thank you!

PR #36, opened 2016-08-22
Add integers in binary format to core with lowercase "b" to maintain functionality of Nordic SDK

dlabun:
Add integers in binary format to core with lowercase "b" to maintain functionality of Nordic SDK.


dlabun:
Turns out Nordic uses "B0" and "B1" within the SDK itself so there's no way implement the exact same binary.h from the SAMD core. The easiest fix I have is to use a lowercase "b" which leaves the Nordic SDK intact and offers this feature with only minor inconvenience to users since they'll need to modify their code slightly.

sandeepmistry:
@dlabun thanks for submitting, we definitely need the uppercase version for compatibility.

If I move the `#include "binary.h"` in Arduino.h down below `#include "delay.h"` it seems ok to compile with the uppercase.

Would you mind trying that and updating the PR if all is good?

dlabun:
@sandeepmistry It still does not compile for me with moving `#include "binary.h"`... Are you compiling against an nRF51 or nRF52 chip? I think the compatibility issue rears its head only with the nRF52 chip as the `nRF52.h` header file as it defines 'B0' and 'B1'

dlabun:
I tracked down B0 and B1 to the nRF52 on board temperature sensor linear function... [http://infocenter.nordicsemi.com/topic/com.nordic.infocenter.nrf52832.ps.v1.1/temp.html](url)

I see two choices:
A) "Break" the onboard temperature sensor and offer direct compatibility with the other Arduino cores
B) "Break" direct compatibility with other the Arduino cores and leave the temperature sensor alone so it can be used in the future

Personally I like option "B" as I could do a PR to the official Arduino cores to use both "b" and "B". This way code could be moved directly from Nordic to Atmel MCUs and there would be minimal changes to go from Atmel to Nordic. It' similar to the situation when you try to run some libraries originally written for the AVR I2C library and it fails on the SAM / SAMD cores.

sandeepmistry:
@dlabun could you please try the https://github.com/sandeepmistry/arduino-nRF5/tree/binary.h branch?

It seems ok when I build for the nRF52DK.

dlabun:
I'm in the process of upgrading my development machine so it might be a day or two before I can test that... But from looking at the code it still runs afoul with the nRF52's onboard temperature sensor as the Nordic SDK files uses B0 and B1 as linearization variables. As long as the temperature sensor is not used there's no issue.

sandeepmistry:
Ok, I'm going to close this in favour of #42.

I think it's an OK comprise for those wanting to use the temperature sensor to include:
```c
#undef B0
#undef B1
```
in their relevant source files, Arduino compatibility should have higher priority :)

@dlabun thanks again for contributing to the project!

PR #37, opened 2016-08-27
Added a menu item to set the low freqency crystal. 

nihaopaul:
this adds and option to set the LFCLK on the nRF52DK which can also be used on custom boards with or without an external 32.xxxkhz crystal, defaults to use external crystal.

cleaned up the PR.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -18,6 +18,7 @@
 menu.chip=Chip
 menu.softdevice=Softdevice
 menu.version=Version
+menu.lfclk=Low Frequency Clock
 
 nRF52DK.name=nRF52 DK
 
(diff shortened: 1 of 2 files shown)


sandeepmistry:
@nihaopaul thanks!

What do you think about adding a "Generic nRF52" variant like the nRF51 version? Would it make pin mapping easier for your custom board?

The LFCLK option could also be useful for the nRF51 generic variant.

nihaopaul:
The naming convention is the same, some testing would need to be done on the nrf51 for both 
`rc_ctiv` and `rc_temp_ctiv` but it seems possible, i think i saw an nrf51-dk in my box i could test on, the change would need to be made in this repo and and the ble repo but totally possible.

sandeepmistry:
@nihaopaul cool, that would be great if you could look into this :)

PR #53, opened 2016-11-06
Add initial Seeed's Arch BLE variant

movingfox:
Modified based on RedBearLab's nRF51822

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -477,3 +477,50 @@ PCA1000X.menu.lfclk.lfrc=RC Oscillator
 PCA1000X.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
 PCA1000X.menu.lfclk.lfsynt=Synthesized
 PCA1000X.menu.lfclk.lfsynt.build.lfclk_flags=-DUSE_LFSYNT
+
+
+Arch_BLE.name=Arch BLE
+
+Arch_BLE.vid.0=0x0d28
+Arch_BLE.pid.0=0x0204
+
+Arch_BLE.upload.tool=sandeepmistry:openocd
+Arch_BLE.upload.interface=cmsisdap
+Arch_BLE.upload.target=nrf51
+Arch_BLE.upload.maximum_size=262144
+
+Arch_BLE.bootloader.tool=sandeepmistry:openocd
+
+Arch_BLE.build.mcu=cortex-m0
+Arch_BLE.build.f_cpu=16000000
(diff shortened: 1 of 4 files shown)


variants/Arch_BLE/variant.cpp
```diff
@@ -0,0 +1,63 @@
...
+  17,
+
+  // D8 - D13
+  18,
+  23,
+  24,
+  25,
+  28,
+  29,
+
```
sandeepmistry: Do you have a link to more info on the board? Can you please add it to the read me?

If it's this board, I think the pin outs are incorrect: http://wiki.seeed.cc/Arch_BLE/

Should be 7 - 15, 23 - 29 here.

boards.txt
```diff
@@ -477,3 +477,50 @@ PCA1000X.menu.lfclk.lfrc=RC Oscillator
...
+Arch_BLE.build.mcu=cortex-m0
+Arch_BLE.build.f_cpu=16000000
+Arch_BLE.build.board=ARCH_BLE
+Arch_BLE.build.core=nRF5
+Arch_BLE.build.variant=Arch_BLE
+Arch_BLE.build.variant_system_lib=
+Arch_BLE.build.extra_flags=-DNRF51
+Arch_BLE.build.float_flags=
+Arch_BLE.build.ldscript=nrf51_xxaa.ld
+
```
sandeepmistry: Are you sure it's the xxac variant?

movingfox:
Hi,

It's the same board. But I followed pinouts of the schematics PDF on the
same wiki page.

https://raw.githubusercontent.com/SeeedDocument/Arch_BLE/master/res/Arch_BLE_v1.0_pdf.pdf

Didn't notice the wiki's board picture doesn't meet the schematics. Would
check the phsical board later to see which one is correct.

sandeepmistry:
@movingfox 

> Would check the phsical board later to see which one is correct.

Thanks, let me know when you've verified what is correct.

movingfox:
Found an official picture of the physical board here.



It seems that the schematics PDF matches the physical board. But I indeed mistook P16 as AREF. It should be P0 actually.

sandeepmistry:
ok, let me know when you've made the adjustments and tested the pin mapping on a physical board.

sandeepmistry:
@movingfox any progress on this?

sandeepmistry:
Closing this for now, due to lack of activity ... please re-open once tested etc.

PR #55, opened 2016-11-06
Add: avr/wdt.h

d00616:
This enables the WDT.

wdt_enable() accepts a timeout in ms. The WDTO_... constants are updated.

wdt_disable() does not work for nRF5. I set a timeout to 36h and configure WDT to pause in sleep mode. Running code from flash with DCDC enabled consumes 133mAh. A CR2032 (250mAh) powered device is reset a maximum of one time in battery life when wdt_disable is used at the beginning.

### DIFF
--- a/cores/nRF5/avr/wdt.h
+++ b/cores/nRF5/avr/wdt.h
@@ -1,0 +1,185 @@
+/* Copyright (c) 2002, 2004 Marek Michalkiewicz
+   Copyright (c) 2005, 2006, 2007 Eric B. Weddington
+   Copyright (c) 2016 Frank Holtz
+   All rights reserved.
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are met:
+
+   * Redistributions of source code must retain the above copyright
+     notice, this list of conditions and the following disclaimer.
+
+   * Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in
+     the documentation and/or other materials provided with the
+     distribution.
+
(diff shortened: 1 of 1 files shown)


sandeepmistry:
Hi @d00616,

Thanks for submitting this!

Have you taken a look to see if there are any more Arduino style libraries for watch dog functionality? It is nice to have something that's compatible with the AVR specific C API's, just wondering if there is a library that wraps them.

> wdt_disable() does not work for nRF5.

Can you please expand on this? Can we make it work?

d00616:
Hello @sandeepmistry ,

i haven't looked for Library. Now i have found a Library here https://github.com/adafruit/Adafruit_SleepyDog but the most Sketches are using avr/wdt.h

For ESP8266 the wdt_ functions are defined in "Esp.h".

> > wdt_disable() does not work for nRF5.
> Can you please expand on this? Can we make it work?

A watchdog timer can only started but not stopped. There is no STOP task. Resetting the mask in NRF_WDT->RREN doesn't work. So there is only an option to set the interval to maximum.

sandeepmistry:
What do you think about forking Adafruit's SleepDog and adding support inside that library?

d00616:
SleepDog is a good candidate to implement some watchdog functionality, but i don't wont fork a lot of libraries.

My plan is to port MySensors to nRF5. To make it uncomplicated to use, i implement most functionality is not provided by arduino-nRF5 into MySensors hardware abstraction layer.

sandeepmistry:
Closing this for now, the SAMD core does not include `avr/wdt.h`

PR #56, opened 2016-11-07
Add: util/atomic.h for Cortex CPU

d00616:
The atomic.h file I have taken from Teensy and tested with nRF51 and nRF52.

### DIFF
--- a/cores/nRF5/util/atomic.h
+++ b/cores/nRF5/util/atomic.h
@@ -1,0 +1,71 @@
+/*
+* This is port of Dean Camera's ATOMIC_BLOCK macros for AVR to ARM Cortex M3 
+* v1.0
+* Mark Pendrith, Nov 27, 2012.
+*
+* From Mark:
+* >When I ported the macros I emailed Dean to ask what attribution would be
+* >appropriate, and here is his response:
+* >
+* >>Mark,
+* >>I think it's great that you've ported the macros; consider them
+* >>public domain, to do with whatever you wish. I hope you find them >useful .
+* >>
+* >>Cheers!
+* >>- Dean 
+*/
+
+#ifdef __arm__ 
+#ifndef _CORTEX_M3_ATOMIC_H_
+#define _CORTEX_M3_ATOMIC_H_
+
+static __inline__ uint32_t __get_primask(void) \
+{ uint32_t primask = 0; \
+  __asm__ volatile ("MRS %[result], PRIMASK\n\t":[result]"=r"(primask)::); \
+  return primask; } // returns 0 if interrupts enabled, 1 if disabled
+
+static __inline__ void __set_primask(uint32_t setval) \
+{ __asm__ volatile ("MSR PRIMASK, %[value]\n\t""dmb\n\t""dsb\n\t""isb\n\t"::[value]"r"(setval):);
+  __asm__ volatile ("" ::: "memory");}
+
+static __inline__ uint32_t __iSeiRetVal(void) \
+{ __asm__ volatile ("CPSIE i\n\t""dmb\n\t""dsb\n\t""isb\n\t"); \
+  __asm__ volatile ("" ::: "memory"); return 1; }   
+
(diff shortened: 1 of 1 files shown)


cores/nRF5/Print.cpp
```diff
@@ -37,6 +37,16 @@ size_t Print::write(const uint8_t *buffer, size_t size)
...
+// for printf compatibility
+extern "C" {
+__attribute__((weak))
+int _write(int file, char *buffer, int size)
+{
+  Serial.write(buffer, size);
+	return 0;
+}
+}
+
```
sandeepmistry: This shouldn't be here?

cores/nRF5/util/atomic.h
```diff
@@ -0,0 +1,71 @@
...
+* >>I think it's great that you've ported the macros; consider them
+* >>public domain, to do with whatever you wish. I hope you find them >useful .
+* >>
+* >>Cheers!
+* >>- Dean 
+*/
+
+#ifdef __arm__ 
+#ifndef _CORTEX_M3_ATOMIC_H_
+#define _CORTEX_M3_ATOMIC_H_
```
sandeepmistry: Does this work on both nRF51 (M0) and nRF52 (M4)?

d00616: I have tested with both MCU. Here is my test sketch. The "counter" is increased circa once a second until 10 seconds. Then a 5 seconds delay is executed in an "ATOMIC" block.

```
#include "nrf.h"
#include <util/atomic.h>

uint16_t counter = 0;

void setup() {
  Serial.begin(38400);

  NVIC_SetPriority(RTC0_IRQn, 15);
  NVIC_ClearPendingIRQ(RTC0_IRQn);
  NVIC_EnableIRQ(RTC0_IRQn);
  NRF_RTC0->PRESCALER = 32;
  NRF_RTC0->EVTENSET = RTC_EVTENSET_COMPARE0_Msk;
  NRF_RTC0->INTENSET = RTC_INTENSET_COMPARE0_Msk;
  NRF_RTC0->CC[0] = 500;
  NRF_RTC0->TASKS_START = 1;
}

void loop() {
  ATOMIC_BLOCK(ATOMIC_FORCEON)
  {
     if (millis()>10000) {
      delay(5000);
     }
  }
  Serial.print("Counter: ");
  Serial.println(counter);
  Serial.print("RTC: ");
  Serial.println(NRF_RTC1->COUNTER);
  Serial.print("Millis: ");
  Serial.println(millis());
  delay(300);
}

extern "C" void RTC0_IRQHandler(void) { counter++; NRF_RTC0->EVENTS_COMPARE[0] = 0; NRF_RTC0->CC[0] = NRF_RTC0->COUNTER + 500; } 
```

cores/nRF5/Print.cpp
d00616: Sorry. I have removed this change.

sandeepmistry:
@d00616 thanks for the info.

I'm still on the fence about including this. The header file is useful, however it's not really part of the standard Arduino API, so using it makes the code less portable.

d00616:
@sandeepmistry This code comes per default with Arduino for AVR and Teensy. There are a lot of sketches using ATOMIC_BLOCK macro. Supporting it helps to make Sketches portable from AVR/Teensy to nRF5. There are other useful definitions in the "utils" folder comming with AVR arduino.

sandeepmistry:
@d00616 yes, but it's not part of the official [SAMD](https://github.com/arduino/ArduinoCore-samd) and [SAM](https://github.com/arduino/Arduino/tree/master/hardware/arduino/sam) cores. Plus they are not documented in the [Arduino Reference docs](https://www.arduino.cc/en/Reference/HomePage).

So overall, only 1 out of 3 of the official Arduino cores have them (AVR), and it's only in AVR because it's bundled with the AVR toolchain.

sandeepmistry:
Closing this for now, the SAMD core does not include `util/atomic.h`.

Issue #71, opened 2016-12-24
Tools are slow to install


dlabun:
This conversation was started at the end of #57 regarding the slowness of the core's tools to install.

dlabun:
@sandeepmistry I ran a test with a modified fork, Using a Zip version of the tools will unpack in about 30 seconds while the existing bzip2 takes 3 minutes (This is with dual Xeons at 3.4Ghz so results will vary). Is Mac OS X capable of dealing with Zip files for the tools?

sandeepmistry:
@dlabun interesting!

> Is Mac OS X capable of dealing with Zip files for the tools?

Yes, the Arduino IDE uses the same Java code to extract archives.


The slowness might also be related to the compression level I selected.

In any case, do you have a size comparison for the files? Could you also prepare a PR for the JSON file and share links to the zip files you made?

dlabun:
@sandeepmistry On Windows the ZIP is ~120MB while the bzip2 is ~101MB. The download will be slightly bigger but the installation performance is significantly better. 

ARM only offers the Windows version as a ZIP so the Linux and Mac versions need to be repackaged first. How do you feel about pushing another tools release so the ZIPs accessible for testing?

sandeepmistry:
> How do you feel about pushing another tools release so the ZIPs accessible for testing?

@dlabun yes, let's go ahead. Do you have time to prepare the zip files and upload to the [`tools` release section](https://github.com/sandeepmistry/arduino-nRF5/releases/tag/tools)? I can test on OS X.

dlabun:
I can have the new ZIPs ready pretty fast... Do you want me to edit the existing tools release or start a tools v2 release?

sandeepmistry:
Let's start with just editing the existing tools (basically v0.2.0, but with zips for arm-gcc). Just to avoid changing too many things at once.

If this goes well we can make a v0.3.0 release afterwards.

dlabun:
ZIPs are up along with a text file with the check sums...  The files structures should be identical to the original versions.

sandeepmistry:
@dlabun nice! I see the zips there!

I think we should make a staging JSON (`package_nRF5_boards_index-staging.json`) in the `docs` folder for testing, then we can copy it over to the current one when ready. I'll try to do this tomorrow, but if you have time to work on it before I start let me know.

dlabun:
I may be able to get to it tonight, I'll let you know if I am able.

[...]

PR #72, opened 2016-12-25
Sort board names in boards.txt

d00616:
I have ordered the board names in boards.txt and added the missing Waveshare board into README.md


sandeepmistry:
@d00616 thanks!

I think once we figure out #71, then we can tag a new release. What do you think?

d00616:
@sandeepmistry I agree.

PR #75, opened 2017-01-07
correction nRF5FlashSoftDevice

d00616:
I have modified the nRF5FlashSoftDevice.java file, because your username is hardcoded in it. This prohibits using a unmodified file in my own package of arduino-nRF5.

In my opinion at the moment there is no need to check the name, because the tool fails if a programmer without a softdevice.txt file is selected.

I have added a check to test if the chosen programmer have to capability to upload a boot loader. 

My plan is to write a boot loader with OTA flash functionality which dosn't depend on accepting the NRF SDK license. Maybe somebody implementing a BLE support for http://mynewt.apache.org/pages/ble/

In my opinion this tool needs to replaced with a "SDK and SoftDevice" downloader. This allows to remove file in https://github.com/sandeepmistry/arduino-nRF5/tree/master/cores/nRF5/SDK with non free copyright. The SoftDevice can flashed via "burn bootloader" option from "Tools" menu.

### DIFF
--- a/extras/ide-tools/build.sh
+++ b/extras/ide-tools/build.sh
@@ -1,7 +1,19 @@
 #!/bin/sh -x
 
-IDE_LIB_PATH=/Applications/Arduino.app/Contents/Java
+# search for arduino IDE: OSX, Linux:opt, Debian, EPEL
+for IDE_LIB_PATH in /Applications/Arduino.app/Contents/Java /opt/arduino/lib /usr/share/arduino/lib/ /usr/share/arduino; do
+	if [ -e "${IDE_LIB_PATH}/pde.jar" ]; then
+		if [ -f "${IDE_LIB_PATH}/arduino-core.jar" ]; then
+			ARDUINO_CORE="arduino-core.jar"
+		else
+			ARDUINO_CORE="core.jar"
+		fi
+		break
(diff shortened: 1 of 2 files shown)


dlabun:
If you want to use a Nordic soft device you have accept their licensing terms, there's no way around it. Your only other choice is to not use the built in wireless features if you object to their terms.

d00616:
@dlabun That's correct. I mean to download SDK and SoftDevice via Tools menu after accepting the license. Without this download there is no SoftDevice binary available to flash.

dlabun:
I've been using a modified version of the core that uses the Nordic flashing tools instead of Sandeep's Java based tool. The upside is that it merges the soft device and your code so you don't need to worry about getting the soft device uploaded. The down side is that it only works with Segger J-Link. I'm not sure how the response would be to using closed sourced flashing tools controlled Nordic / Segger.

sandeepmistry:
@d00616 I'm not keen on merging this, as with the change there is no validation for it the user selected a valid programmer for flashing.

d00616:
@sandeepmistry If you select an invalid programmer there is a message like: "Unsupported programmer! 'buspirate'".

The changed build-script supports various Linux distributions.

PR #85, opened 2017-01-25
Add initial ng-beacon variant

urish:
This is a small nrf51 based board that I created, modeled after the Angular logo. I plan to publish a blog post about it next week, and mention arduino-nRF5 and arduino-BLEPeripheral as the recommended way to develop for it. If you could merge this PR before the post is published, that would be awesome :-)


dlabun:
I looked at your pull request and it looks like you duplicated the generic variant. Are you sure that is going work with your board?

urish:
Yes, tested it before I submitted it. The main differences are the pin
mappings for I2C, UART and SPI.

dlabun:
The other question I had was could you provide a pin out or schematic for the board... I couldn't find one to  compare to the variant configuration.

urish:
Sure, here is the pin out for the PTR5518 module that is used by the board:

https://devzone.nordicsemi.com/attachment/82ad87b813c019c429d744fef8121a32

Uri

dlabun:
I'm reluctant to say this variant is okay, the variant information does not match the module you're using. For example the variant says there are 32 pins when this module only has 11. Unless Sandeep wants to accept it as is, I would recommend correcting the variant to match the module.

urish:
Good note on the number of pins! I'm totally open to make changes, are
there any other issues that you see here?

On Jan 25, 2017 6:50 PM, "dlabun" <notifications@github.com> wrote:

I'm reluctant to say this variant is okay, the variant information does not
match the module you're using. For example the variant says there are 32
pins when this module only has 11. Unless Sandeep wants to accept it as is,
I would recommend correcting the variant to match the module.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/sandeepmistry/arduino-nRF5/pull/85#issuecomment-275163069>,
or mute the thread

variants/ng-beacon/variant.h
```diff
@@ -0,0 +1,103 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Lesser General Public License for more details.
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#ifndef _VARIANT_GENERIC_
+#define _VARIANT_GENERIC_
```
sandeepmistry: This should be updated accordingly.

variants/ng-beacon/variant.h
```diff
@@ -0,0 +1,103 @@
...
+
+/** Master clock frequency */
+#ifdef NRF52
```
sandeepmistry: I think you can remove this ...

sandeepmistry:
I agree with @dlabun's comments, there are more pins in the variant than are exposed on the board.

urish:
Thanks @sandeepmistry, I will update the PR now

urish:
PR updated :)

dlabun:
@sandeepmistry looks a lot better now

sandeepmistry:
@dlabun thanks for reviewing!

urish:
Thank you very much gentlemen! Any chance for a new release? I would really love to get this released for the ng-beacons blog post I am publishing this week :)

sandeepmistry:
Me and @dlabun are trying to sort out #71 before the next release, hopefully it will get sorted out by the end of this weekend, but no guarantees.

Maybe you can suggest using the instructions to install from git, and then update the blog post when a new release has been cut?

PR #87, opened 2017-01-30
Fixing millis() and micros()

austin1611:
Previously there were two issues for the millis() and micros() functions on nRF52:

1) in delay.c  : The counter is 24 bits, so when it overflows we should add 2^24 to the count, not 2^32.

2) in wiring.c : The overflow interrupt was not enabled (or not enabled properly).


austin1611:
Here is an example arduino script which will reveal the error:

```
// millis test

uint32_t timeInMilliseconds = 0;

// changes to be made
// 1) Enable the overflow interrupt (in wiring.c, insert near line 36)
//      NRF_RTC1->INTENSET = 2; 
// 2) Modify the overflow shifting (in delay.c) to use 24 bits not 32 (replace lines 32,39)
//    uint64_t ticks = (uint64_t)((uint64_t)overflows << (uint64_t)24) | (uint64_t)(NRF_RTC1->COUNTER);

void setup() 
{
  Serial.begin(115200);
}

void loop() {


  // cheater way to get an overflow/interrupt sooner than 512 seconds:
  if (timeInMilliseconds > 5000){
    NRF_RTC1->TASKS_TRIGOVRFLW = 1;
  }

  timeInMilliseconds = millis();
  Serial.println(timeInMilliseconds);
  delay(100);
  
  if (timeInMilliseconds > 512000)
  {
    Serial.println("Successfully passed 2^24 ticks of the RTC (~511.9 seconds)");
    if (timeInMilliseconds > 600000)
    {
      Serial.println("but the overflow count wasn't added to the total time");  
    }
    while(1==1){}
  }

}
```

cores/nRF5/wiring.c
```diff
@@ -34,6 +34,7 @@ void init( void )
   NRF_CLOCK->LFCLKSRC = (uint32_t)((CLOCK_LFCLKSRC_SRC_RC << CLOCK_LFCLKSRC_SRC_Pos) & CLOCK_LFCLKSRC_SRC_Msk);
   NRF_CLOCK->TASKS_LFCLKSTART = 1UL;
 
+  NRF_RTC1->INTENSET = 2;
```
sandeepmistry: What do you think about using the `RTC_INTENSET_OVRFLW_Msk` constant here instead of 2?

Declared:

 * https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/SDK/components/device/nrf51_bitfields.h#L5465
 * https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/SDK/components/device/nrf52_bitfields.h#L11483

austin1611:
Sandeep, the bitmask should be fine. 
I updated wiring.c - see if that's what you expected.

cores/nRF5/wiring.c
```diff
@@ -34,6 +34,7 @@ void init( void )
   NRF_CLOCK->LFCLKSRC = (uint32_t)((CLOCK_LFCLKSRC_SRC_RC << CLOCK_LFCLKSRC_SRC_Pos) & CLOCK_LFCLKSRC_SRC_Msk);
   NRF_CLOCK->TASKS_LFCLKSTART = 1UL;
 
+  NRF_RTC1->INTENSET = RTC_INTENSET_OVRFLW_Msk;
   NRF_RTC1->PRESCALER = 0;
   NRF_RTC1->EVTENSET = offsetof(NRF_RTC_Type, EVENTS_OVRFLW);
   NRF_RTC1->INTENSET = offsetof(NRF_RTC_Type, EVENTS_OVRFLW);
```
sandeepmistry: Things are really busted now, this whole block needs to be replaced with:

```c
  NRF_RTC1->PRESCALER = 0;
  NRF_RTC1->EVTENSET = RTC_INTENSET_OVRFLW_Msk;
  NRF_RTC1->INTENSET = RTC_EVTEN_OVRFLW_Msk;
  NRF_RTC1->TASKS_START = 1;
```

As ` NRF_RTC1->INTENSET` get overwritten with `offsetof(NRF_RTC_Type, EVENTS_OVRFLW);` which is completely wrong. After waiting over 512 seconds, the `millis()` value wrapped with the following sketch:

```arduino
void setup() {
  Serial.begin(9600);

  Serial.println("NRF_RTC1->INTENSET");
  Serial.println(NRF_RTC1->INTENSET, HEX);
}

void loop() {
  Serial.println(millis());
  delay(1000);
}
```

@austin1611 could you please make this change and test again? After that, we are good this merge this!

Thanks.

sandeepmistry:
@austin1611 thanks!

PR #90, opened 2017-02-09
Config gpio as pinreset

bojh:
Nordic does provide a standard mapping of a reset functionality on P.21 which is in most boards the reset button. To support the RESET button in a standard Arduino way we have to provide a c compile flag CONFIG_GPIO_AS_PINRESET.


bojh:
Opps - something went wrong

PR #97, opened 2017-02-19
LF CLCK source cleanup

sandeepmistry:
This sets the LF CLK source for all variants except the generic variants. I've only left the menu items for generic boards, if someone wants to tests an alternative LF CLK source for a non-generic board, they should just use the generic variant instead - as this keeps the boards.txt cleaner.

@dlabun please review, after this is merged, we can finally make a new v0.3.0 release.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -87,6 +87,8 @@ Blend2.build.extra_flags=-DNRF52
 Blend2.build.float_flags=-mfloat-abi=hard -mfpu=fpv4-sp-d16
 Blend2.build.ldscript=nrf52_xxaa.ld
 
+Blend2.build.lfclk_flags=-DUSE_LFXO
+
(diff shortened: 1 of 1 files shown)


dlabun:
Looks good code wise... But maybe there should be a note in the readme that the clock menu option has been removed for all but the generic? It may prevent some issues in the future.

dlabun:
Merging as is... Will include a note in the next release about the change to the LF CLCK changes.

sandeepmistry:
@dlabun thanks for merging. Adding a note to the read me is a great idea! As it was released part of the 0.2.0 release.

dlabun:
@sandeepmistry A changelog may be in order at this point, what do you think?

sandeepmistry:
@dlabun agreed.

PR #100, opened 2017-02-25
Allow access to the 8 analog inputs (VS 6 previously).

honnet:
Hi @sandeepmistry!
1st off, congrats for this amazing repo, I've been following your work for a while and it's always a pleasure to read and use your code.

About this pull request, it's a simple modification that allows accessing to all the analog pins of the nRF52.
The Arduino environment doesn't allow it by default but several forks (such as Teensy) have more analog inputs and a similar approach.
Let me know if this PR is not formatted as it should, any feedback or suggestion would be more than welcome.
Thanks again for this great project,
Cedric ;)

PS: As an example, the modification is used it here:
https://github.com/honnet/nRF52AnalogTx/blob/master/nRF52AnalogTx.ino#L14

### DIFF
--- a/variants/nRF52DK/variant.cpp
+++ b/variants/nRF52DK/variant.cpp
@@ -38,13 +38,15 @@ const uint32_t g_ADigitalPinMap[] = {
   24,
   25,
 
-  // A0 - A5
+  // A0 - A7
   3,
   4,
   28,
   29,
   30,
   31,
+  5, // AIN3 (P0.05)
+  2, // AIN0 (P0.02) / AREF
 
   // SDA, SCL
   26,
@@ -52,8 +54,5 @@ const uint32_t g_ADigitalPinMap[] = {
 
   // RX, TX
   8,
-  6,
-
-  // AREF
-  2
+  6
 };
--- a/variants/nRF52DK/variant.h
+++ b/variants/nRF52DK/variant.h
@@ -32,9 +32,9 @@ extern "C"
 #endif // __cplusplus
 
 // Number of pins defined in PinDescription array
-#define PINS_COUNT           (25u)
+#define PINS_COUNT           (26u)
 #define NUM_DIGITAL_PINS     (20u)
-#define NUM_ANALOG_INPUTS    (6u)
+#define NUM_ANALOG_INPUTS    (8u)
 #define NUM_ANALOG_OUTPUTS   (0u)
 
 // LEDs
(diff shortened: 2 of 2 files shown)


sandeepmistry:
Hey @honnet,

Thanks for the positive feedback and the PR!

Are the A6 and A7 broken out on the dev board?

honnet:
Yes, and they're connected as this:
A6 = P0.02 = AIN0 = AREF
A7 = P0.05 = AIN3

P0.02 and 05 are visible on this schematics:
http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52%2Fdita%2Fnrf52%2Fdevelopment%2Fnrf52_dev_kit%2Fhw_drawings_v1.0.html&cp=2_2_2_5_0

And the nrf52 pinout is available here:
http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Fpin.html&cp=2_2_0_3&anchor=pin_assign

I had a few of opportunities to work with Nordic MCUs (including http://twiz.io and http://remora.carpenoctem.cc), it's impressive how your work is making us save time...
Thanks again!

sandeepmistry:
Cool, thanks for the info + links!

I'm thinking the A6 and A7 pins in variant.cpp should be consecutive then, like right after the A5 entries. Then we'll need to bump the SDA, SCL, RX and TX values in variant.h accordingly. What do you think?

honnet:
I thought the same...
But I tried to make this commit as simple as possible and as A6 was already there as AREF (=AIN0), I just added A7 beside.
Should I do it anyway and make another pull request?
(I could also use push -f)

sandeepmistry:
Makes sense :)

Either is fine (new PR or force push).

We'll be able to do something like this still, this way:

```c
for (int aPin = A0; aPin < (A0 + NUM_ANALOG_INPUTS); aPin++) {
  int blah = analogRead(aPin);

  // do something with blah ...
}
```

honnet:
Sounds good!
That was the idea here:
https://github.com/honnet/nRF52AnalogTx/blob/master/nRF52AnalogTx.ino#L14
But it will be better ;)

honnet:
OK, it should be good now (I squashed your suggestions and did push -f):
https://github.com/honnet/arduino-nRF5/commit/master

Also, I tested it here:
https://github.com/honnet/nRF52AnalogTx/blob/master/nRF52AnalogTx.ino#L63-L64

Basically, I moved 2 AINx pins before I2C and UART pins so I incremented all every following index by 2.
Also, AIN0 (AKA P0.02 or AREF) is at the end of the analog pins because of its double role (AIN + AREF).

The squashed commit seems clear enough but if you want to see the latest step before the squash, it's on another branch :
https://github.com/honnet/arduino-nRF5/commit/detailed_steps

dlabun:
@sandeepmistry Nope, looks good to me.

honnet:
Thanks guys, I hope to get to collaborate more soon ;)

PR #112, opened 2017-03-02
first pass at tinyble variant

jacobrosenthal:
Working PR, no rush to merge, comments accepted.

Theres only 4 pins available on the tinyble so Im duping them across all digital, analog, spi, and i2c. Also theres an accelerometer on board but Im not defining that in the variant .. not sure if I should? Have to take a look at how whatever MPU6050 library works


variants/TinyBLE/variant.h
```diff
@@ -0,0 +1,98 @@
...
+
+#include "WVariant.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif // __cplusplus
+
+// Number of pins defined in PinDescription array
+#define PINS_COUNT           (32u)
```
sandeepmistry: Any reason to expose all the pins even though they aren't broken out?

jacobrosenthal:
Thanks for the code review!

I dont know.. didnt know it mattered I guess?

So it should be all externally interface-able pins. Ive noticed Serial is defined on say, nrf dongle, which only goes through the segger and thus isnt actually broken out and probably shouldnt be incldued right? Nor SPI and I2C?

jacobrosenthal:
Also whenever I get specific the blink doesnt blink.. Is there an ordering here?

sandeepmistry:
Sorry, not PIN_COUNT, but `NUM_DIGITAL_PINS`

I think you should be able to loop through 0 - NUM_DIGITAL_PINS and send HIGH or LOW on some external pins. So those exclude the non-broken out SPI or I2C pins, but could include them if they were broken out to headers.

jacobrosenthal:
Based on my last discussion with sandeep I set NUM_DIGITAL_PINS to 8 (3 LEDS, 1 Button, 4 exposed pins)

boards.txt
```diff
@@ -439,6 +439,48 @@ nRF51Dongle.menu.softdevice.s130.upload.maximum_size=151552
 nRF51Dongle.menu.softdevice.s130.build.extra_flags=-DNRF51 -DS130 -DNRF51_S130
 nRF51Dongle.menu.softdevice.s130.build.ldscript=armgcc_s130_nrf51822_{build.chip}.ld
 
+
+TinyBLE.name= TinyBLE
+
+TinyBLE.upload.tool=sandeepmistry:openocd
+TinyBLE.upload.protocol=cmsis-dap
```
sandeepmistry: This depends on #114 right?

variants/TinyBLE/variant.h
```diff
@@ -0,0 +1,99 @@
...
+//3 LEDS, 1 Button, 4 exposed pins
+#define NUM_DIGITAL_PINS     (8u)
```
sandeepmistry: I think this is good now, to confirm the digital pins are:

21, 22, 23, 17, 3, 4, 5, 6?

jacobrosenthal: Yep

boards.txt
jacobrosenthal: I would argue our default selection has never worked, everyone has manually selected up until now, so it doesnt depend on it, but rather will be useful when 114 lands.

sandeepmistry:
I'll merge this once #114 is merged.

Issue #111, opened 2017-03-02
Document programmer to use with standard boards


jacobrosenthal:
Its currently a guessing game to know that the nrf51 dongle uses a jlink programmer, but mbed boards use a cmsis dap, etc, etc.

It would be good at the very least to compile a list of these. Anyone want to drop a comment which programmer to use with their favorite boards?

Then where to put that, we list the supported boards in the readme. .It could go there. This is clumsy, but we could also try to smoosh it into the board name in the arduino menu item?

jacobrosenthal:
I see a upload.interface defined as jlink and cmsis-dap in boards.txt but not sure that it does anything. Anyone know what this is for or a link for how arduino uses it? Can it select the appropriate programmer by default?

sandeepmistry:
> Anyone want to drop a comment which programmer to use with their favorite boards?

Most of the boards I use have built-in CMSIS DAP or JLINKS, I use a STLink clone with my BluzDK.

> Anyone know what this is for or a link for how arduino uses it? Can it select the appropriate programmer by default?

See [`tools.openocd.upload.pattern="{path}/{cmd}" {upload.verbose} -f interface/{upload.interface}.cfg -c "{upload.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"`](https://github.com/sandeepmistry/arduino-nRF5/blob/master/platform.txt#L119)

Here's my understanding, if the boards.txt entry specifies `<board>.upload.protocol=` the `tools.openocd.upload.pattern` is used, other wise `tools.openocd.program.pattern` is used. You can also select `Sketch -> Upload Using Programmer` for force using the programmer.

jacobrosenthal:
Super Interesting. None of the boards use upload.protocol. they do have upload.interface, and a bunch of them are defined as cmsisdap, which if it was supposed to be fed into openocd would be wrong (cmsis-dap.cfg is the setting)

Any idea if upload.interface is valid/used? 

I also needed 
TinyBLE.upload.setup_command=transport select swd;

dlabun:
All of my stuff is programmed with a jlink onboard chip or my external jlink

jacobrosenthal:
Also of note, softdevice upload still requires a programmer to be set sadly. It seems like this could/should be altered to use the program pattern to be overridden from boards.txt

sandeepmistry:
[...]

PR #114, opened 2017-03-04
consolidates to using .protocol instead of .interface and correct defines

jacobrosenthal:
I believe upload.interface is incorrect, and upload.protocol is the proper command
cmsisdap was mispelled
and cmsisdap will probably? always need swd option

See #111

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -116,11 +116,11 @@ tools.openocd.cmd.windows=bin/openocd.exe
 
 tools.openocd.upload.params.verbose=-d2
 tools.openocd.upload.params.quiet=-d0
-tools.openocd.upload.pattern="{path}/{cmd}" {upload.verbose} -f interface/{upload.interface}.cfg -c "{upload.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
+tools.openocd.upload.pattern="{path}/{cmd}" {upload.verbose} -f interface/{upload.protool}.cfg -c "{upload.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
 
 tools.openocd.program.params.verbose=-d2
 tools.openocd.program.params.quiet=-d0
-tools.openocd.program.pattern="{path}/{cmd}" {program.verbose} -f interface/{program.interface}.cfg -c "{program.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
+tools.openocd.program.pattern="{path}/{cmd}" {program.verbose} -f interface/{program.protocol}.cfg -c "{program.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
 
 tools.openocd.erase.params.verbose=-d3
 tools.openocd.erase.params.quiet=-d0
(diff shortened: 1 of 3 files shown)


sandeepmistry:
@jacobrosenthal does this change also require a change to `tools.openocd.upload.pattern="{path}/{cmd}" {upload.verbose} -f interface/{upload.interface}.cfg -c "{upload.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
` line in platform.txt?

Other than that it makes sense, I forget why I decided to go for the empty value path when I started.

jacobrosenthal:
Which means program.interface is also suspect here. 

I presume that needs to be changed to something like program.protocol, but I dont thats a real thing yet either, and seems related to https://github.com/sandeepmistry/arduino-nRF5/issues/111#issuecomment-283830118

Perhaps just leave that be until we resolve that PR upstream?

jacobrosenthal:
So .interface is actually a totally fine custom key, however there are already a bunch of 'standards' for this with board.upload.protocol and both programmer.protocol, programmer.program.protocol 

(why both??, theyre upstream like that so Ill just leave them and duplicate them like they often do upstream https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/programmers.txt)

This PR consolidates to using .protocol instead of .interface, and attempts to populate all .protocol fields, and corrects the typo cmsisdap to cmsis-dap

jacobrosenthal:
OK taking one more shot at clarifying this, if only for myself

Arduino defines 
https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification

>  •	upload a sketch to the target board (using a bootloader preinstalled on the board)
>  •	program a sketch to the target board using an external programmer
>  •	erase the target board's flash memory using an external programmer
>  •	burn a bootloader into the target board using an external programmer
> 

We don't have a boot loader, so our program and upload are identical except upload comes from board definition (defined in the core) and program comes from programmer definition (and is selected via menu by the user)

When the user hits upload we call [uploadUsingPreferences](https://github.com/arduino/Arduino/blob/master/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java#L65)
If the user forced "using programmer" or the board defined board.upload.protocol is empty (which ours IS now as we use currently use .interface) then [uploadUsingProgrammer](https://github.com/arduino/Arduino/blob/master/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java#L89) is actually used.
- boardPreferences go into prefs
- programmerPrefs merged ON TOP of prefs
- Then prefs is used to fill program.pattern

So currently our upload.pattern has NEVER been used. and our programmerPrefs have been wiping out our boardPreferences up til now

This changes that to fix this be aligning to upload.protocol, which WOULD skip uploadUsingProgrammer
- boardPreferences goes into prefs
- Then prefs is used to fill upload.pattern


Only tangentially related:

For uploading soft device we utilize the bootloader pathway, [uploader.burnbootloader](https://github.com/arduino/Arduino/blob/master/arduino-core/src/cc/arduino/packages/uploaders/SerialUploader.java#L352)
- boardPreferences goes into prefs
- programmerPrefs are merged ON TOP of prefs
- toolPrefs are merged ON TOP of prefs
- Then prefs is used to fill bootloader.pattern

platform.txt
```diff
@@ -116,16 +116,16 @@ tools.openocd.cmd.windows=bin/openocd.exe
 
 tools.openocd.upload.params.verbose=-d2
 tools.openocd.upload.params.quiet=-d0
-tools.openocd.upload.pattern="{path}/{cmd}" {upload.verbose} -f interface/{upload.interface}.cfg -c "{upload.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
+tools.openocd.upload.pattern="{path}/{cmd}" {upload.verbose} -f interface/{upload.protool}.cfg -c "{upload.setup_command}" -f target/{upload.target}.cfg -c "program {{{build.path}/{build.project_name}.hex}} verify reset; shutdown;"
```
sandeepmistry: `upload.protool` - typo?

Should be `upload.protocol`?

sandeepmistry:
Once the typo is corrected I think this is good to go.

@dlabun what do you think?

dlabun:
My only question is how easily can this change be rolled back? Seems like a real easy way to brick the entire core by messing with the programming methods.

jacobrosenthal:
This change just fixes our default bootloading selection, nothing more low
level than that.

And all of these devices have hardware bootloading via jlink or cmsis, so
bricking seems not to too much a worry for me.

OwenBrotherwood:
Can this be pulled so @carlosperate can get his PR done: he is waiting for this one, so he can fix up conflicts.

dlabun:
@jacobrosenthal @sandeepmistry Have either of you tested these changes yet?

sandeepmistry:
Tested this with the micro:bit:

```
/Users/smistry/Library/Arduino15/packages/sandeepmistry/tools/openocd/0.10.0-dev.nrf5/bin/openocd -d2 -f interface/cmsis-dap.cfg -c transport select swd; -f target/nrf51.cfg -c program {{/var/folders/5d/pmvtw7p579n5npm6p421qf540000gn/T/arduino_build_830030/Blink.ino.hex}} verify reset; shutdown; 
```

Looks good, merging!

PR #115, opened 2017-03-06
Add NFR51 Smart Beacon Kit variant

joaopedrovbs:
I'm using the **NRF51 Smart Beacon Board**. As I didn't find direct support to it I added it to the list! Checked all the `I/Os` and made `LED_BUILTIN` work as well. I also added some flags to the `Generic NRF51822` if someone wants to use it with some of the original **Nordic SDK Libraries**.


boards.txt
```diff
@@ -228,7 +228,7 @@ Generic_nRF51822.build.board=GENERIC
 Generic_nRF51822.build.core=nRF5
 Generic_nRF51822.build.variant=Generic
 Generic_nRF51822.build.variant_system_lib=
-Generic_nRF51822.build.extra_flags=-DNRF51
+Generic_nRF51822.build.extra_flags=-DNRF51 -DBOARD_PCA20006 -DNRF51822
```
joaopedrovbs: Ohh yes, that was there just for Debug while I built the other Board! thanks for the advice!

dlabun:
@sandeepmistry What do you think?

boards.txt
```diff
@@ -228,7 +228,7 @@ Generic_nRF51822.build.board=GENERIC
 Generic_nRF51822.build.core=nRF5
 Generic_nRF51822.build.variant=Generic
 Generic_nRF51822.build.variant_system_lib=
-Generic_nRF51822.build.extra_flags=-DNRF51
+Generic_nRF51822.build.extra_flags=-DNRF51 -DNRF51822
```
sandeepmistry: The extra ` -DNRF51822` can be removed here.

boards.txt
```diff
@@ -261,6 +261,51 @@ Generic_nRF51822.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
...
+
+Beacon_PCA20006.bootloader.tool=sandeepmistry:openocd
+
+Beacon_PCA20006.build.mcu=cortex-m0
+Beacon_PCA20006.build.f_cpu=16000000
+Beacon_PCA20006.build.board=GENERIC
+Beacon_PCA20006.build.core=nRF5
+Beacon_PCA20006.build.variant=PCA20006
+Beacon_PCA20006.build.variant_system_lib=
+Beacon_PCA20006.build.extra_flags=-DNRF51 -DBOARD_PCA20006 -DNRF51822
```
sandeepmistry: We can remove the `-DBOARD_PCA20006 -DNRF51822` args, the IDE should generate an equivalent for the `BOARD_PCA20006` - see the verbose build logs.

boards.txt
```diff
@@ -261,6 +261,51 @@ Generic_nRF51822.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
...
+Beacon_PCA20006.menu.chip.xxac=32 kB RAM, 256 kB flash (xxac)
+Beacon_PCA20006.menu.chip.xxac.build.chip=xxac
+
+Beacon_PCA20006.menu.softdevice.none=None
+Beacon_PCA20006.menu.softdevice.none.softdevice=none
+Beacon_PCA20006.menu.softdevice.s110=S110
+Beacon_PCA20006.menu.softdevice.s110.softdevice=s110
+Beacon_PCA20006.menu.softdevice.s110.softdeviceversion=8.0.0
+Beacon_PCA20006.menu.softdevice.s110.upload.maximum_size=151552
+Beacon_PCA20006.menu.softdevice.s110.build.extra_flags=-DNRF51 -DS110 -DNRF51_S110 -DBOARD_PCA20006 -DNRF51822
```
sandeepmistry: Same applies here.

boards.txt
```diff
@@ -261,6 +261,51 @@ Generic_nRF51822.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
...
+Beacon_PCA20006.menu.softdevice.s110.build.ldscript=armgcc_s110_nrf51822_{build.chip}.ld
+
+Beacon_PCA20006.menu.softdevice.s130=S130
+Beacon_PCA20006.menu.softdevice.s130.softdevice=s130
+Beacon_PCA20006.menu.softdevice.s130.softdeviceversion=2.0.1
+Beacon_PCA20006.menu.softdevice.s130.upload.maximum_size=151552
+Beacon_PCA20006.menu.softdevice.s130.build.extra_flags=-DNRF51 -DS130 -DNRF51_S130 -DBOARD_PCA20006 -DNRF51822
```
sandeepmistry: Same for this ...

boards.txt
```diff
@@ -261,6 +261,51 @@ Generic_nRF51822.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
...
+Beacon_PCA20006.menu.softdevice.s130.build.ldscript=armgcc_s130_nrf51822_{build.chip}.ld
+
+Beacon_PCA20006.menu.lfclk.lfxo=Crystal Oscillator
+Beacon_PCA20006.menu.lfclk.lfxo.build.lfclk_flags=-DUSE_LFXO
+Beacon_PCA20006.menu.lfclk.lfrc=RC Oscillator
+Beacon_PCA20006.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
+Beacon_PCA20006.menu.lfclk.lfsynt=Synthesized
+Beacon_PCA20006.menu.lfclk.lfsynt.build.lfclk_flags=-DUSE_LFSYNT
```
sandeepmistry: Does the board have a crystal oscillator built in? If so we can remove these options.

variants/PCA20006/variant.h
```diff
@@ -0,0 +1,103 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Lesser General Public License for more details.
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#ifndef _VARIANT_GENERIC_
+#define _VARIANT_GENERIC_
```
sandeepmistry: This should be updated accordingly.

variants/PCA20006/variant.h
```diff
@@ -0,0 +1,103 @@
...
+
+/** Master clock frequency */
+#ifdef NRF52
```
sandeepmistry: I think the NRF52 section doesn't apply.

variants/PCA20006/variant.h
```diff
@@ -0,0 +1,103 @@
...
+#include "WVariant.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif // __cplusplus
+
+// Number of pins defined in PinDescription array
+#define PINS_COUNT           (32u)
+#define NUM_DIGITAL_PINS     (32u)
```
sandeepmistry: Does the board have 32 external pins broken out? If not this to be the same as the broken out pins. The BluzDK is probably a better variant to base things on.

variants/PCA20006/variant.h
```diff
@@ -0,0 +1,103 @@
...
+#define PIN_A4               (5)
+#define PIN_A5               (6)
+
+static const uint8_t A0  = PIN_A0 ;
+static const uint8_t A1  = PIN_A1 ;
+static const uint8_t A2  = PIN_A2 ;
+static const uint8_t A3  = PIN_A3 ;
+static const uint8_t A4  = PIN_A4 ;
+static const uint8_t A5  = PIN_A5 ;
+#ifdef NRF52
```
sandeepmistry: NRF52 section can probably be removed.

joaopedrovbs:
So, @sandeepmistry the pins are not all broken out in this board, but in this case the pins are already labeled. If I declare them as only 15 pins, it will cause a bad experience for the user that has to find out that for example, the pin 24 at his board is in fact pin 14 on software... How should I proceed to comply with the project? Make it 32 pins and populate the array with a big number on the pins that don't exist?

sandeepmistry:
Would you be able to share a detailed photo of the board?

joaopedrovbs:
Board ImageHere you go!

sandeepmistry:
So the broken out pins are: 0, 1, 3, 5, 9, 11, 20, 21, 24? 

Does the board have a external crystal connected?

If so I agree, things would be confusing if we try to make then consecutive. @dlabun @jacobrosenthal what do you think?

joaopedrovbs:
Yes, those are the pins. The board has a 16 MHz external crystal. While making my tests i configured the other pins to a high number just while we talk through this... Also I had to make a fix in order to make I2C work with my MPU9250 should I open another PR to talk about it or use this one?

sandeepmistry:
@joaopedrovbs just push the commits to the branch you used for this PR please.

joaopedrovbs:
@sandeepmistry The full story is, we were having some serious trouble using the MPU9250 IMU, the I2C bytes usually got out of sync and this made the sensor unrecognizable and the values unusable and constant. The speed used was 400kHz and could not be lowered. We started to jump in the wire implementation to look for the problem. We compared it to the Nordic implementation and added some functions from there like clear and enabled the use of shorcuts to stop having conflicts with the S130 SoftDevice. That solved the problems for us.

jacobrosenthal:
@joaopedrovbs I think you and @sandeepmistry miscommunicated, you wont want i2c changes in the variant PR

joaopedrovbs:
Hello @jacobrosenthal, I've seen now @carlosperate PR and I think that fixes the problem I was having in a simpler way. I'll make sure to roll back on this branch and leave I2C untouched

boards.txt
```diff
@@ -228,7 +228,11 @@ Generic_nRF51822.build.board=GENERIC
 Generic_nRF51822.build.core=nRF5
 Generic_nRF51822.build.variant=Generic
 Generic_nRF51822.build.variant_system_lib=
+<<<<<<< Updated upstream
+Generic_nRF51822.build.extra_flags=-DNRF51 -DNRF51822
+=======
 Generic_nRF51822.build.extra_flags=-DNRF51
+>>>>>>> Stashed changes
```
carlosperate: Looks like you've got an unresolved git conflict here.

joaopedrovbs: Yes, I'm working on that right now, fixing and rolling back on that changes

joaopedrovbs: @carlosperate Do you have some advice on what to do on the issue with the pins that aren't broken out? For now I'm leaving them with a big value like: 
`const uint32_t g_ADigitalPinMap[] = {
  0,
  1,
  2,
  3,
  50}`

Is that ok, as I can't use -1 or 0 ?

carlosperate: I think `-1` would be used for pins that "do not exists", but still need to be in the array for other pins to not take that value.
So if you have a header with 4 pins, 0 to 3, but pin 2 in the header is marked with something that is not a microcontroller GPIO (so maybe ground or Vcc), you still need a `g_ADigitalPinMap[]` with 4 values, so then it'd be `g_ADigitalPinMap[] = { 0, 1, -1, 50}`. It doesn't matter what the nordic pin values are (as long as they are real pins), so you can add them in whatever order makes sense for your board breakout, since that's the array that does the mapping between "arduino pin" and "nordic pin". So in the example above `digitalRead(3)` maps to pin 50 if the nRF.
Did that make sense? does that answer your question?

joaopedrovbs: Take a look at the picture posted above and my discussion with @sandeepmistry, the board from Nordic has just pins `0, 1, 3, 5, 9, 11, 20, 21, 24` broken out. for a better user experience I think it's good to maintain pin 24 as `digitalRead(24)` for example.. Should I populate the other numbers on the array, with `(uint32_t)-1`, or their real number, even though they won''t be used?

carlosperate: Yes, if they are already labelled in the board with the nrf pin, then is better to keep the nrf and arduino pin numbering the same.
Then I believe -1 is the value used in other variants for that.

carlosperate: I would only leave their real number if the pins are actually connected to something else. So even if they are broken out, are they any other components were those pins could be wired to?

There is at least an RGB LED that the user would still want to use, no? Is there anything else in the board?

joaopedrovbs: There's an RGB LED and 2 Buttons, their pins are also with Nordic Pin Number, as it is on the boards "datasheet".

sandeepmistry:
I think this is good to merge now.

@jacobrosenthal @carlosperate what do you think?

carlosperate:
LGTM 👍 
And for reference, schematic and more info can be found in http://infocenter.nordicsemi.com/pdf/nRF6930_Beacon_Ref_Design_UG_v1.1.pdf

PR #123, opened 2017-03-24
Servo library

carlosperate:
Tested it in a BBC micro:bit, it does the job, is not quite ready yet to be merged but can be reviewed. There are a few things I am not sure about and the timings seem a bit off.
I'll add review comments to individual lines of the PR code to ask a few questions.

This has been based from the SAMD version of the library, and I've also looked at the redbear implementation, should a note be added somewhere indicating this? A README maybe?

### DIFF
--- a/libraries/Servo/Servo.cpp
+++ b/libraries/Servo/Servo.cpp
@@ -1,0 +1,305 @@
+/*
+  Copyright (c) 2015 Arduino LLC. All right reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
+#include <Arduino.h>
+#include <Servo.h>
+
+// Converts microseconds to timer tick and vice versa
+#define usToTicks(_us)     ((TMR_FREQ / 1000000L) * (_us))
+#define ticksToUs(_ticks)  ((unsigned)_ticks / (TMR_FREQ / 1000000L))
+
+#define TRIM_DURATION  5              // compensation ticks to trim adjust for digitalWrite delays
+
+static servo_t servos[MAX_SERVOS];    // static array of servo structures
+
+static uint8_t ServoCount = 0;        // the total number of attached servos
+
+// Index for the servo being pulsed for each timer (or -1 if refresh interval)
+static volatile int8_t currentServoIndex[_Nbr_16timers];
+#define REFRESH_INTERVAL_PERIOD            (-1)
+
+// Timer controlling this servo
+#define SERVO_INDEX_TO_TIMER(_servo_nbr)   ((timer16_Sequence_t)(_servo_nbr / SERVOS_PER_TIMER))
+// Index of the servo on this timer
+#define SERVO_INDEX_TO_CHANNEL(_servo_nbr) (_servo_nbr % SERVOS_PER_TIMER)
+// Servo index by timer and channel
+#define SERVO_INDEX(_timer, _channel)      ((_timer * SERVOS_PER_TIMER) + _channel)
+// Servo structure by timer and channel
+#define SERVO(_timer, _channel)            (servos[SERVO_INDEX(_timer,_channel)])
+
+/************ static functions common to all instances ***********************/
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+inline void Servo_Handler(timer16_Sequence_t timer, NRF_TIMER_Type *nrfTimer, uint32_t ccReg)
+{
+  // Clear the interrupt event flag
+  nrfTimer->EVENTS_COMPARE[ccReg] = 0;
+
+  if (currentServoIndex[timer] == REFRESH_INTERVAL_PERIOD) {
+    // Refresh interval completed so reset the timer
+    nrfTimer->TASKS_CLEAR = 1;
+    nrfTimer->CC[ccReg] = 0;
+  } else {
+    if (SERVO_INDEX(timer, currentServoIndex[timer]) < ServoCount &&
+        SERVO(timer, currentServoIndex[timer]).Pin.isActive == true) {
+      // It's an active channel so pulse it low
+      NRF_GPIO->OUTCLR = digitalPinToBitMask(SERVO(timer, currentServoIndex[timer]).Pin.nbr);
+    }
+  }
+
+  // Select the next servo controlled by this timer
+  currentServoIndex[timer]++;
+  if (SERVO_INDEX(timer, currentServoIndex[timer]) < ServoCount &&
+      currentServoIndex[timer] < SERVOS_PER_TIMER) {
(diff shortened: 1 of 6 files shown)


libraries/Servo/Servo.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if (SERVO(timer, currentServoIndex[timer]).Pin.isActive == true) {
+      // It's an active channel so pulse it high
+      NRF_GPIO->OUTSET = digitalPinToBitMask(SERVO(timer,currentServoIndex[timer]).Pin.nbr);
+    }
+    // Get the counter value
+    unsigned int ticks = SERVO(timer, currentServoIndex[timer]).ticks;
+    nrfTimer->CC[ccReg] += ticks + (ticks >> 6);
+  } else {
+    // Finished all channels so wait for the refresh period to expire before starting over
+    // Allow a few ticks to ensure the next timer event is not missed
```
carlosperate: I am not quite sure exactly why this `+ 4` is done, I've seen it in multiple implementations of the servo library for different platforms, although I assume it's been carried over from the original Arduino implementation.
Is it just to just to give us some head room in case the counter (which keeps running within the interrupt) hits the REFRESH_INTERVAL value? I guess that way we don't miss it and gets trigger very soon after the interrupt?

libraries/Servo/ServoTimers.h
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * NRF5 Only definitions
+ * ---------------------
+ */
+
+// Uncomment to select allocatable timers, 3 and 4 are only available in nRF52
+//#define _useTimer0
+//#define _useTimer1
+#define _useTimer2
```
carlosperate: I've only really tested with this timer at the moment, I'll try to test the rest soon, but I see no reason why they shouldn't work unless there are in used already by something else.

libraries/Servo/ServoTimers.h
```diff
@@ -0,0 +1,122 @@
...
+#include "nrf.h"
+
+/**
+ * NRF5 Only definitions
+ * ---------------------
+ */
+
+// Uncomment to select allocatable timers, 3 and 4 are only available in nRF52
+//#define _useTimer0
+//#define _useTimer1
```
carlosperate: Timer 1 is currently being used in the `analogWrite` implementation for the nRF51 (not nRF52).
Is that a permanent implementation? Should I add a `#error` if this is uncommented in the nRF51 platform?

libraries/Servo/ServoTimers.h
```diff
@@ -0,0 +1,122 @@
...
+
+#include "nrf.h"
+
+/**
+ * NRF5 Only definitions
+ * ---------------------
+ */
+
+// Uncomment to select allocatable timers, 3 and 4 are only available in nRF52
+//#define _useTimer0
```
carlosperate: I'm not familiar with the SoftDevices configuration, I vaguely remember seeing mentions of timer 0. If that's the case perhaps we should remove this as an option altogether?

libraries/Servo/ServoTimers.h
```diff
@@ -0,0 +1,122 @@
...
+//#define _useTimer1
+#define _useTimer2
+//#define _useTimer3
+//#define _useTimer4
+
+
+// Timer IRQ priority should take in consideration other interrupts
+#define TMR_PRIORITY                 1
```
carlosperate: I've just left the priority set to 1 during testing, but perhaps it should be placed a bit lower.
I think at the moment GPIOTE is also priority 1, the Wire I2C is set to priority 2, UART to priority 3, the analogWrite TIMER1 at priority 3 as well, and the RTC all the way down to 15.

libraries/Servo/examples/Sweep/Sweep.ino
```diff
@@ -0,0 +1,62 @@
...
+  }
+  delay(3000);
+  for (i = 180; i >= 0; i -= 1) { // goes from 180 degrees to 0 degrees
+    myservo.write(i);
+    myservo1.write(i);
+    myservo2.write(i);
+    delay(50);
+  }
+}
+
```
carlosperate: I'll probably replace this file with the default Sweep sketch, been using this code for general testing.

carlosperate:
There is one very odd thing that I've noticed, which is that it seemed that the timer was significantly off. But rather than being a deterministic offset (there are a couple of branches in the interrupt code, but each path should still be deterministic), it seems to be proportional to the length of the timer, at around 2% more or less.

That is the reason why I added the following extra ticks to the counter:

```
nrfTimer->CC[ccReg] += ticks + (ticks >> 6);
...
nrfTimer->CC[ccReg] = (uint32_t)usToTicks(REFRESH_INTERVAL) + (REFRESH_INTERVAL >> 5);
```

Since the timer clock is derived from the HFCLK external crystal, and it's not likely the clock is that innacurate I am open to suggestions, as I am not quite sure what else could be causing it.
Also, I've only tested it in the micro:bit, so I am not sure if this will have the same problem in other nRF5 boards.

libraries/Servo/Servo.cpp
```diff
@@ -0,0 +1,305 @@
+/*
+  Copyright (c) 2015 Arduino LLC. All right reserved.
```
sandeepmistry: I've been adding a new line: https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/wiring_analog_nRF51.c#L3

libraries/Servo/ServoTimers.h
sandeepmistry: I think RTC0 is used by the soft device, RTC1 is used by the core for delay(), millis(): https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/delay.c#L59

libraries/Servo/ServoTimers.h
sandeepmistry: Agreed, or just remove the comments :)

libraries/Servo/ServoTimers.h
carlosperate: Are you sure it's just the RTC0? I haven't read the soft device documentation, but looking at this it sounds like it might also use TIMER0:
[cores/nRF5/SDK/components/softdevice/s132/headers/nrf_nvic.h#L76-L110](https://github.com/sandeepmistry/arduino-nRF5/blob/8747fa2d5947dd3cde1d94abf796dd2c17781177/cores/nRF5/SDK/components/softdevice/s132/headers/nrf_nvic.h#L76-L110)

libraries/Servo/ServoTimers.h
carlosperate: I assume it can still be used with the nRF52 boards? I was wondering why there were different implementations in that respect, in case it might be temporary  and both end up using or not using the TIMER1.

carlosperate: I've just noticed the nRF52 boards have a dedicated PWM peripheral, so this is probably the reason there are two implementations. I can edit the macros to allow use of the TIMER1 only on the nRF52 boards, as it is not used otherwise.

dlabun:
@carlosperate Is the servo library ready to go?

carlosperate:
Not really, I was hoping to get some feedback in the outstanding questions I added to the review.
My main concern is the odd timing offset, which works on my test system, but I don't really have other nRF5 boards to test if the same would apply to them, only micro:bits.

On that note, I've noticed the same % error in the RTC timer for the delay operation. So if I do a `delay(2)` or `delay(200)` it seems to be about that 2-3% off in both cases.

carlosperate:
As it currently stands the RTC1 timer has priority 15 and this library has priority 1 (which is something that should probably be changed to something lower, as mentioned in one of the code comments).
Even if it was the other way around, I should have clarified that this time offset I'm seeing is only in one direction, so I'm seeing the `delay(x)` trigger about 2-3% quicker than expected. And the same with the servo if I didn't add the extra ticks as mentioned in https://github.com/sandeepmistry/arduino-nRF5/pull/123#issuecomment-289109780.

sandeepmistry:
Closing this for now due to lack of activity. Please let us know if/when this is ready for review and testing.

PR #125, opened 2017-03-25
Set microbit build.board value

carlosperate:
Tiny PR to correct the build.board name so we can identify the micro:bit board in the preprocessor.


sandeepmistry:
@carlosperate thanks for correcting this!

PR #126, opened 2017-03-26
Add AVR interrupt sei() and cli() macros

carlosperate:
Some sketches use these AVR specific interrupt macros (sei and cli), so they can be added into `cores/nRF5/avr/interrupt.h` for compatibility.

### DIFF
--- a/cores/nRF5/avr/interrupt.h
+++ b/cores/nRF5/avr/interrupt.h
@@ -17,7 +17,9 @@
 */
 
 /*
-  Empty file.
   This file is here to allow compatibility with sketches (made for AVR)
   that includes <AVR/interrupt.h>
 */
+
+#define sei() __enable_irq()
+#define cli() __disable_irq()


sandeepmistry:
@carlosperate thanks for submitting this!

I'd prefer to leave this out as the SAMD core doesn't provide this: https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/avr/interrupt.h

Any reason not to use:
 * https://www.arduino.cc/en/Reference/Interrupts
 * https://www.arduino.cc/en/Reference/NoInterrupts

carlosperate:
I completely agree that any sketch should be using the `interrupts` and `noInterrupts()` macro, and it is an easy enough change, but only if you know what to edit and where it is.
After all, if the new core includes an empty header to allow compilation for sketches importing `<AVR/interrupt.h>`, then it seems like a good place to add a working version of what might be used from this header file.

Also, the other files in https://github.com/arduino/ArduinoCore-samd/tree/master/cores/arduino/avr do contain a few entries for compatibility reasons, maybe this is something that could be included in all of them?

sandeepmistry:
Depends, if the other official Arduino cores (SAMD, 101, SAM) contained this, I'm be more include it.

@dlabun @jacobrosenthal what are your thoughts?

jacobrosenthal:
Yeah I think the arduino-ey way of interrupts/nointerrupts is already implemented in
https://github.com/sandeepmistry/arduino-nRF5/blob/dc53980c8bac27898fca90d8ecb268e11111edc1/cores/nRF5/Arduino.h

dlabun:
In my experience interrupts never translate from AVR to any of the other cores... I'd say exclude.

carlosperate:
> In my experience interrupts never translate from AVR to any of the other cores... I'd say exclude.

This is just to enable and disable interrupts, normally used to execute atomic operations, nothing to do with the interrupts themselves.

> Yeah I think the arduino-ey way of interrupts/nointerrupts is already implemented in ...

And this is the absolute right way of doing it, but really, that's not the point. The reason we have to include a mock  `/avr/interrupt.h` file is precisely for compatibility with naughty sketches that are using the AVR libraries directly, so this was just a way to extend this compatibility a little bit further.

Nevertheless I do understand why this might be out of the scope of this Arduino core, and happy to close it.
Thanks for the consideration.

jacobrosenthal:
Which libs are being naughty? Cant we harass them and link to this issue?

carlosperate:
I wasn't talking about any specific libraries or sketches, but the fact that this file is required precisely because of that (we have the same comment in this repo, but it likely originated from https://github.com/arduino/ArduinoCore-sam/commit/157d3d26dda85c3bfc216d59a6332f05ccc52be0).

A simple search within github will get you thousands of hits, although it is true that most of these are too specific to the atmel architecture where they are just not portable to other cores.
https://github.com/search?l=Arduino&q=sei%28%29+arduino&type=Code&utf8=%E2%9C%93

The good libraries will have their AVR specific code guarded by defines doing conditional compilation, but if we trusted that this was the case for all (or all that matter), then we wouldn't need to have this file in the repository any more, wouldn't we?

PR #128, opened 2017-03-27
Set CMSIS-DAP as the default uploader for the micro:bit

carlosperate:
Since the micro:bit comes with a DAPLink uC, we can configure CMSIS-DAP as the default uploader.


carlosperate:
Looks like this PR might be in small conflict with https://github.com/sandeepmistry/arduino-nRF5/pull/114 (consolidating the board.protocol and board.interface), so whichever is merged first will affect the other.

sandeepmistry:
@carlosperate thanks!

We can rebase #114 and update.

PR #131, opened 2017-03-29
Fix RTC1 EVTENSET/INTENSET and RTC1_IRQHandler

d00616:
- Correction of EVTENSET/INTENSET


carlosperate:
Since that bit of code is being updated, maybe it'd be a good thing to also apply enable it more explicitly?

```cpp
NRF_RTC1->INTENSET = (uint32_t)(RTC_INTENSET_OVRFLW_Enabled << RTC_INTENSET_OVRFLW_Pos);
NRF_RTC1->EVTENSET = (uint32_t)(RTC_EVTEN_OVRFLW_Enabled << RTC_EVTEN_OVRFLW_Pos);
```

sandeepmistry:
@d00616 thanks!

@carlosperate I think your suggestion is equivalent:

```
#define RTC_INTENSET_OVRFLW_Msk (0x1UL << RTC_INTENSET_OVRFLW_Pos) /*!< Bit mask of OVRFLW field. */

:#define RTC_EVTEN_OVRFLW_Msk (0x1UL << RTC_EVTEN_OVRFLW_Pos) /*!< Bit mask of OVRFLW field. */
```

carlosperate:
Oh yes, it is definitely equivalent, it is just more explicit to set the "enable bit" in the "right position" rather than relaying on the mask value to be the same as the enable value (which actually should be the case in all registers).
It was just a suggestion to follow the same convention with all the register macros.

PR #132, opened 2017-04-01
Add SDA and SCL pin constants to variant definitions.

carlosperate:
So that arduino sketches can refer directly to `SDA` and `SCL` pins.

I've only added these variables to variants with `PIN_WIRE_SDA` and `PIN_WIRE_SCL` already defined, although one or two already had them.


jacobrosenthal:
@sandeepmistry asked to check against the samd variant as thats the newest implementation from arduin and thus the gold standard and these are indeed in there. Thanks for the PR!
https://github.com/arduino/ArduinoCore-samd/blob/95963e6ff76257e4816bcd62bdd3914de73ef656/variants/arduino_zero/variant.h#L154

Issue #134, opened 2017-04-05
Nordic PAN 56: TWI module lock -up


carlosperate:
I've been trying to debug some issues I am having with the I2C driver (I'll submit an unrelated PR for that soon), and googling around I've encountered quite a few mentions of this Product Anomaly Notice # 56 for nRF51822 and nRF51422.

The following conditions will make the TWI lock up: 
* Disabling the TWI module while the TWI slave clock-stretches an ongoing read or write sequence.
* Triggering the RESUME task too early after receiving RXDRDY event. Firmware triggers the RESUME task 
within the same CPU clock cycle as the ACK bit is finished sending from the TWI master

This only affects the silicon v2.4 or earlier (circa 2014), and while I assume v3.x is fairly common these days, going through my boards I've found at least one that would be affected. Since the software reset workaround is not that bad, it's probably not unreasonable to include it.

http://infocenter.nordicsemi.com/pdf/nRF51822-pan_v2.4.pdf

dlabun:
It sounds like it'd take some work to really get to the point of locking up the TWI module... Personal opinion, I don't think this really needs to be built in but it's good to know about for the fringe cases.

jacobrosenthal:
@carlosperate  do you know which variants we already define might be susceptible to this?

carlosperate:
Well, I assume most of the nRF51 variants use the nRF51822 or nRF51422, so technically most if not all of the nRF51?  It's a matter when "when" were the boards built, as newer batches might contained the newer silicon version with the bugfix.

sandeepmistry:
Closed via https://github.com/sandeepmistry/arduino-nRF5/issues/137.

PR #137, opened 2017-04-06
Fix nRF51 I2C read operation

carlosperate:
I've been having issues reading I2C data using the TWI driver. In the smallest test case
I was reading 2 individual bytes one after another in a loop and the data returned from the 'Wire.read()' was in the incorrect order.

The operation itself is very simple, send a write command selecting the register to be read and then read a single byte back. The operation should read register `0` with value `0x40` and then register `1` with value `0x00`, as shown in the image below. The bus looks exactly the same with and without this patch, but without it the software was returning the value of register `0` as `0x00` and register `1` as `0x40`.

image

The fix is achieved by enabling during read the shortcut between the "byte boundary event" and the "suspend task" for each read byte (so when it receives a byte it triggers the hw suspend until we manually resume operation in the read loop), except for the last one, which has the byte boundary event and stop task shortcut. This is the same approach followed by the NRF sdk implementation.

Apart from that TwoWire::requestFrom() wasn't really returning the number of bytes read, so that is fixed as well, together with a clamp on input argument `quantity`, so that the number of bytes to read cannot be larger than the size of the rx ring buffer.

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -28,9 +28,7 @@
 
 #include "RingBuffer.h"
 
-#define BUFFER_LENGTH 32
-
- // WIRE_HAS_END means Wire has end()
+// WIRE_HAS_END means Wire has end()
 #define WIRE_HAS_END 1
 
 class TwoWire : public Stream
@@ -91,7 +89,7 @@ class TwoWire : public Stream
     // RX Buffer
     RingBuffer rxBuffer;
 
-    //TX buffer
+    // TX buffer
     RingBuffer txBuffer;
     uint8_t txAddress;
 
--- a/libraries/Wire/Wire_nRF51.cpp
+++ b/libraries/Wire/Wire_nRF51.cpp
@@ -36,7 +36,7 @@ TwoWire::TwoWire(NRF_TWI_Type * p_twi, IRQn_Type IRQn, uint8_t pinSDA, uint8_t p
   this->_IRQn = IRQn;
   this->_uc_pinSDA = g_ADigitalPinMap[pinSDA];
   this->_uc_pinSCL = g_ADigitalPinMap[pinSCL];
-  transmissionBegun = false;
+  this->transmissionBegun = false;
 }
 
 void TwoWire::begin(void) {
@@ -55,7 +55,7 @@ void TwoWire::begin(void) {
                                 | ((uint32_t)GPIO_PIN_CNF_DRIVE_S0D1       << GPIO_PIN_CNF_DRIVE_Pos)
                                 | ((uint32_t)GPIO_PIN_CNF_SENSE_Disabled   << GPIO_PIN_CNF_SENSE_Pos);
 
-  _p_twi->FREQUENCY = TWI_FREQUENCY_FREQUENCY_K100;
+  _p_twi->FREQUENCY = (TWI_FREQUENCY_FREQUENCY_K100 << TWI_FREQUENCY_FREQUENCY_Pos);
   _p_twi->ENABLE = (TWI_ENABLE_ENABLE_Enabled << TWI_ENABLE_ENABLE_Pos);
   _p_twi->PSELSCL = _uc_pinSCL;
   _p_twi->PSELSDA = _uc_pinSDA;
@@ -83,7 +83,7 @@ void TwoWire::setClock(uint32_t baudrate) {
     frequency = TWI_FREQUENCY_FREQUENCY_K400;
   }
 
-  _p_twi->FREQUENCY = frequency;
+  _p_twi->FREQUENCY = (frequency << TWI_FREQUENCY_FREQUENCY_Pos);
   _p_twi->ENABLE = (TWI_ENABLE_ENABLE_Enabled << TWI_ENABLE_ENABLE_Pos);
 }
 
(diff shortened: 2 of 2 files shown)


libraries/Wire/Wire.h
```diff
@@ -28,9 +28,7 @@
 
 #include "RingBuffer.h"
 
-#define BUFFER_LENGTH 32
```
carlosperate: This is a left over from the AVR Wire implementation and is also incorrect for this core, as the size of the ring buffers used is 64.

cheeta1:
@carlosperate  thanks for the correction i was also facing this  i2c reading issue for too long.

jacobrosenthal:
@joaopedrovbs has some i2c work in this branch as well
 https://github.com/sandeepmistry/arduino-nRF5/pull/115 I like that this is in its own PR :) @joaopedrovbs anything missing from here? @carlosperate anything you like in his?

carlosperate:
One thing that that PR has is the software workaround for https://github.com/sandeepmistry/arduino-nRF5/issues/134, but I am a bit worried about the code included in that PR as it looks like a straight copy-paste from the nordic sdk, and we all know about their funny licensing.

joaopedrovbs:
@carlosperate That's really the case, I got stuck for some days trying to make it work and went straight to Nordic Implementation checking what they had and ours didn't, I'm new to this nordic world, what you think is the best to do?

carlosperate:
I'd be great if you could confirm if this version of the Wire library fixes the issue you had, or if it works as well as the version in your PR does. If so I think it'd be better to merge this one, which tackles only the problem we had reading I2C, while maintaining the rest of the driver implementation the same.

And then if deemed appropriate by the discussion in #134, we could add `TwoWire::clear()` from #115. At that point we could see how close it is from the sdk or the PAN 56 solution, but we can cross that bridge once we get there.

joaopedrovbs:
So @carlosperate tested your implementation and it failed to read as well. I'll check later the differences between both implementations to see what I can change...

carlosperate:
What were you trying to do and what problems did you have?

joaopedrovbs:
So, the thing I'm work on consists of the small Nordic Beacon Kit, connected to an MPU9250 IMU, the errors I had to fix were pretty much like the ones you showed in this PR, there's a byte that kept reappearing on the bus, on my case 0x20, this would not allow the MPU to find in it's secondary I2C bus the compass... With my code it's working well, as I put this code the IMU didn't begin at all. By this time I can't debug it deeper as i'm without the proper tools for that

reapqin:
@joaopedrovbs i am also working with nrf51 and mpu9250. i try to fix the code as @carlosperate do in wire.h to delete BUFFER_LENGTH 32 line. but the other libraries I2Cdev use this define to do somethine. How to do next ? 
Thanks

libraries/Wire/Wire.h
reapqin: i fixed here in my code as you deleted this line. but another libararies named I2Cdev uses this define. how to do next?

carlosperate: The `SERIAL_BUFFER_SIZE` macro is provided by "RingBuffer.h", which can be used for the original purpose of this macro.
Maybe it'd be better to do `#define BUFFER_LENGTH SERIAL_BUFFER_SIZE `? Which library uses this macro that is not specific to AVR?

carlosperate:
@reapqin I've responded in https://github.com/sandeepmistry/arduino-nRF5/pull/137/files#r111687740

libraries/Wire/Wire.h
reapqin: it's not work yet
I2Cdev come from https://github.com/jrowberg/i2cdevlib
it's provided nrf51 version but it's not for this i think.
i change this lib used nrf51 version
when i compile code it's error: nrf_drv_twi.h: No such file or directory
get this nrf_drv _twi.h will fix this error?
i am beginer thanks

carlosperate: Sounds like it needs the nordic libraries, which this repository doesn't use. You'd probably get better support creating an issue in https://github.com/jrowberg/i2cdevlib

sandeepmistry: I'm ok with removing it, the SAMD core doesn't have this define.

libraries/Wire/Wire_nRF51.cpp
```diff
@@ -55,7 +55,7 @@ void TwoWire::begin(void) {
                                 | ((uint32_t)GPIO_PIN_CNF_DRIVE_S0D1       << GPIO_PIN_CNF_DRIVE_Pos)
                                 | ((uint32_t)GPIO_PIN_CNF_SENSE_Disabled   << GPIO_PIN_CNF_SENSE_Pos);
 
-  _p_twi->FREQUENCY = TWI_FREQUENCY_FREQUENCY_K100;
+  _p_twi->FREQUENCY = (TWI_FREQUENCY_FREQUENCY_K100 << TWI_FREQUENCY_FREQUENCY_Pos);
```
sandeepmistry: This is fine, but a bit weird to shift by 0:

```
#define TWI_FREQUENCY_FREQUENCY_Pos (0UL) /*!< Position of FREQUENCY field. */
```

sandeepmistry:
I haven't test this out on a board, but code wise it looks good.

@carlosperate is there a specific board that experiences this issue that you were using? I think I was using RBL Nano or nR51822 board when working on the Wire lib initially ...

libraries/Wire/Wire_nRF51.cpp
carlosperate: Yeah it's one of those changes that should result in the same compiled code. I prefer it for consistency (all registers are usually set this way), and portability (although this specific file is nRF51 only).

carlosperate:
I've been testing with the BBC micro:bit. I could try to find a library for the on-board accelerometer or magnetometer, could make for another good test that is easy for other people with the same board to replicate.

sandeepmistry:
@carlosperate that would be awesome! I think I have a pretty old rev of the micro:bit here (pre-outside UK availability).

libraries/Wire/Wire_nRF51.cpp
sandeepmistry: Ok, we'll have to do another pass through all the other source in this repo for it to be consistent - but that's another PR :)

sandeepmistry:
I'm going to merge this one for now even though I haven't physically tested on a board.

PR #138, opened 2017-04-07
Remove unused interrupt from nRF51 Wire lib

carlosperate:
As the nRF51 TWI implementation is master only and is not interrupt driven, there is no need to have any of the interrupt code present in this version.
Also, since the constructor signature is already different to the nRF52 version, is not like it could be kept for compatibility reasons.


sandeepmistry:
@carlosperate thanks!

PR #66, opened 2016-12-05
analogRead() fix for nRF52 by setting related gain.

bojh:
Signed-off-by: Joerg Hubschneider <joerg.hubschneider@bosch-si.com>

a modified standard AnalogInput example with helping comment is attached:
[test_AnalogInput.zip](https://github.com/sandeepmistry/arduino-nRF5/files/631001/test_AnalogInput.zip)

PR #142, opened 2017-04-10
Fix ADC prescaler reg config, add prescaler options

carlosperate:
So, during my tests I noticed that my ADC readings were a bit off. Basically on a voltage of around 3.27v where I would expect the ADC value to be close to the maximum of 10-bit range I was instead reading values slightly above 900.

So for default ADC configuration, the first thing to note is the macro used to set `ADC_CONFIG_INPSEL` was assigning a `ADC_CONFIG_RES` value instead of one of the INPSEL ones. This value, happended to be the same as having 1/3 prescaler on the input signal (so in this case my 3.27v input would be scaled to 1/3 of its value before reaching the ADC comparator). The `ADC_CONFIG_REFSEL` value was then set to the internal voltage reference, which is 1.2V, so in essence, together with the prescaler the ADC was configured to read voltages from 0 to 3.6V. I guess this is fine once you know it, but maybe it's not an ideal configuration taking in consideration a lot of the boards have been designed for 3.3V general operation.

Following one of the PRs for the nRF52 I've added local global to contain a prescaler value to be configured as part of the `analogReference()` set-ups. I've fixed some of the invalid comments and added a few more to easily see the values I have set up in this PR, which I do consider be the safest assumption as to what the user would expect, but of course maybe there are better options, so these should be open for discussion.

I'd like to also emphasise that this PR, as it currently stands, is not changing the default ADC behaviour, but I do very strongly suggest to change this default configuration from what it is right now (1.2V internal reference, 1/3 input prescaler), to the `AR_SUPPLY_ONE_THIRD` configuration.

This `AR_SUPPLY_ONE_THIRD` config should set up the reference voltage to be 1/3 VDD and the input voltage to 1/3 as well, therefore having an ADC range from 0 to VDD. In most boards this would be 0 to 3.3V and I do believe that's what most people would expect out of the box.

### DIFF
--- a/cores/nRF5/wiring_analog.h
+++ b/cores/nRF5/wiring_analog.h
@@ -26,7 +26,7 @@ extern "C" {
 #endif
 
 /*
- * \brief SAMD products have only one reference for ADC
+ * \brief nRF51 and nRF52 have different reference options.
  */
 #ifdef NRF52
 typedef enum _eAnalogReference
--- a/cores/nRF5/wiring_analog_nRF51.c
+++ b/cores/nRF5/wiring_analog_nRF51.c
@@ -48,6 +48,7 @@ static struct PWMContext pwmContext[PWM_COUNT] = {
 static int timerEnabled = 0;
 
 static uint32_t adcReference = ADC_CONFIG_REFSEL_VBG;
+static uint32_t adcPrescaling = ADC_CONFIG_INPSEL_AnalogInputOneThirdPrescaling;
 
 static uint32_t readResolution = 10;
 static uint32_t writeResolution = 8;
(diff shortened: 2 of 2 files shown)


sandeepmistry:
I haven't tested it on a board, but code wise this looks good. @dlabun @jacobrosenthal any thoughts on this?

I believe the referenced nRF52 is https://github.com/sandeepmistry/arduino-nRF5/pull/66.

carlosperate:
Does anybody else also agree that changing the default to the `AR_SUPPLY_ONE_THIRD` configuration (not done in this PR) is a better configuration?
That would set it up to 1/3 input prescaler, and 1/3 Vdd as reference, giving an ADC range from 0v to Vdd (3.3V for most boards), instead of the 0V-3.6V it currently measures.

Suschman:
Out of the box without manual configuration i would expect an ADC to use Zero to VDD sampling. ✓

sandeepmistry:
@carlosperate @Suschman that sounds good to me. What are the next steps to get this done?

carlosperate:
Well, if we are happy with this configuration change to start using prescalers we can either merge this PR and I would create a new one just to change the default configuration. Or I could add a new commit to the PR with the default value changed.

sandeepmistry:
@carlosperate merged :) Thanks for the contribution!

carlosperate:
Thanks! I'll get the next PR ready.

Issue #133, opened 2017-04-03
analogRead() should be able to map inputs 0-to-x as A0-to-Ax


carlosperate:
The default Arduino analogRead() can take 2 different values for each ADC input:
1. The "analog pin" value, which goes from 0 to NUM_ANALOG_INPUTS, which in most Arduino Uno compatible boards would be 0 to 5:
    ````arduino
    analogRead(0);
    analogRead(1);
    ...
    analogRead(5);
    ````
2. The pin number, in it's "integer form" or `Ax` global variable:
    ````arduino
    analogRead(A0);
    analogRead(A1);
    ...
    analogRead(A5);
    ````
    Which in boards like the Uno would be the equivalent of:
    ````arduino
    analogRead(14);
    analogRead(15);
    ...
    analogRead(19);
    ````

This is usually done by simply checking the range of the input argument, so in the case of the Uno the `analogRead()` function would have something along the lines of:

```arduino
analogRead(uint8_t pin) {
  if (pin < A0) {
    pin += A0;
  }
}
```

That works well because these boards are configured to have the analog pins mapped immediately after the digital pins and in successive order.

This Arduino core has variants with analog pin mappings that do not follow this convention, and so only "method 2" listed above has been implemented.

This breaks compatibility with any arduino sketches using "method 1", so we should work out a method to map these analog pins in the variant files in a way that allows both methods to work as expected.

sandeepmistry:
@carlosperate I agree, could you submit a PR for this?

Maybe we can have a `#define` for non-compatible variants - the ones with non-Arduino like pin definitions to disable this option. What do you think?

carlosperate:
Maybe I can put something together in a couple of days, depends on availability, and how much work left there might be for the other PRs.

I think we need to do a pin conversion similar to `if (pin < A0) pin += A0;`, but since this would be specific to each board, a macro/function would have to be defined per variant.

So, in essence, if the input is `< NUM_ANALOG_INPUTS`, then we'd apply this "variant dependant" macro/function, and if not then just continue as it is.

sandeepmistry:
I was thinking something super simple like:

```c
#ifndef ARDUINO_NON_STANDARD_ANALOG_PINS
if (pin < A0) {
  pin += A0;
}
#endif
```

[...]

PR #150, opened 2017-04-21
Identify analogRead() input type and set right pin

carlosperate:
This first commit is only for the micro:bit board, but it is meant to be a starting point to agree on the correct implementation and then extend it to the rest of the variants.

A description of the problem can be found in https://github.com/sandeepmistry/arduino-nRF5/issues/133 and this initial solution was described in the comment https://github.com/sandeepmistry/arduino-nRF5/issues/133#issuecomment-293106851.

### DIFF
--- a/cores/nRF5/wiring_analog_nRF51.c
+++ b/cores/nRF5/wiring_analog_nRF51.c
@@ -119,6 +119,16 @@ uint32_t analogRead( uint32_t ulPin )
   uint32_t resolution;
   int16_t value;
 
+#ifdef ARDUINO_ADC_NON_STANDARD_PINS
+  ulPin = _analogChannelPin(ulPin);
+#elif NUM_ANALOG_INPUTS > 0
+  if (ulPin < A0) {
+    ulPin += A0;
+  }
+#else
+  return 0;
+#endif
+
   if (ulPin >= PINS_COUNT) {
     return 0;
   }
--- a/variants/BBCmicrobit/variant.h
+++ b/variants/BBCmicrobit/variant.h
@@ -40,7 +40,6 @@ extern "C"
 // LEDs
 #define PIN_LED                 (13)
 #define LED_BUILTIN             PIN_LED
-// #define
 
 // Buttons
 #define PIN_BUTTON_A            (5)
@@ -64,6 +63,23 @@ static const uint8_t A3  = PIN_A3 ;
 static const uint8_t A4  = PIN_A4 ;
 static const uint8_t A5  = PIN_A5 ;
 #define ADC_RESOLUTION    10
+
+// Non standard distribution of Analog Pins requires a lookup table
+#define ARDUINO_ADC_NON_STANDARD_PINS
+__inline__ uint32_t _analogChannelPin(uint32_t pin) __attribute__((always_inline));
+__inline__ uint32_t _analogChannelPin(uint32_t pin) {
+  switch (pin) {
+    case 0:  return PIN_A0;
+    case 1:  return PIN_A1;
+    case 2:  return PIN_A2;
+    case 3:  return PIN_A3;
+    case 4:  return PIN_A4;
+    case 5:  return PIN_A5;
+    case 10: return PIN_A5;
+    // Otherwise non-analog pin so return out of range pin to exit early
+    default: return PINS_COUNT;
+  }
+}
 
 /*
  * Serial interfaces


carlosperate:
Btw, the main reason I haven't named the function `analogInputToDigitalPin()` as done in the AVR core macro ([GitHub/Arduino/hardware/arduino/avr/variants/standard/pins_arduino.h#L30](https://github.com/arduino/Arduino/blob/89e3c5290f4fc95db4159f348d6b4e326d415e64/hardware/arduino/avr/variants/standard/pins_arduino.h#L30)) is because the macro returns `-1` for non-valid inputs, and since all the types involved here are unsigned I though it would be easier to follow without casting.
Nevertheless signed to unsigned conversion is well-defined, and casting `-1` (not matter the size) to uint32 would result in `0xFFFFFFFF`, and also cause the same early exit when `(ulPin >= PINS_COUNT)`.

Not sure if it's better to keep it as it is, or change it to `analogInputToDigitalPin()`.

sandeepmistry:
@carlosperate nice!

I would prefer using `analogInputToDigitalPin`. I'm open to making all the pin mapping arrays in the variants to signed if it helps.

carlosperate:
Yeah, after reading my previous message again I think it is better to use `analogInputToDigitalPin()`. If the SAMD core version of `analogRead` has unsigned integers in its signature I think it's fine keeping it as it is. I'll explicitly add the cast to the return value so that it's clear a `-1` would end up as `0xFFFFFFFF`, and it we'll be good to go.

carlosperate:
Damn, so looking at it again I've changed my mind once more and there are a couple of issues I didn't originally realised.

Okay, the first thing is that, as it currently stands, the `_analogChannelPin()` inline function from this PR should perhaps stay with that (or similar) name instead of using `analogInputToDigitalPin` because it does a bit more than what the `analogInputToDigitalPin()` macro does.
For each "ADC channel" value (i.e. 0 to 5 in an Arduino Uno) it returns the equivalent digital pin, so that's the same; but it also returns a valid digital pin for an input value that corresponds to a valid analog pin (so in this case for the micro:bit pin 10 is ADC channel 5, and so `_analogChannelPin(10)` does returns `10` instead of `-1`).

The other thing is that, if the "digital pin value" for an an "ADC pin" overlaps with the ADC channel range (so let's say ADC channel 0 to 5 are digital pins 2 to 7) then there is no way to detect when the input argument is supposed to represent the channel vs. when it is meant to represent the digital input value.

So depending on the implementation the following loop could give unexpected results:
```c
#define PIN_A0               (2)
#define PIN_A1               (3)
#define PIN_A2               (4)
#define PIN_A3               (5)
#define PIN_A4               (6)
#define PIN_A5               (7)
#define NUM_ANALOG_INPUTS    (6u)

for (size_t i = 0; i < NUM_ANALOG_INPUTS; i++) {
  Serial.println(_analogChannelPin(i))
}
```

For the following lookup table, which prioritises channels, we would give the result below:

```c

_analogChannelPin(uint32_t pin) {
  switch (pin) {
    case 0:  return PIN_A0;
    case 1:  return PIN_A1;
    case 2:  return PIN_A2;
    case 3:  return PIN_A3;
    case 4:  return PIN_A4;
    case 5:  return PIN_A5;
    case PIN_A4: return PIN_A4;
    case PIN_A5: return PIN_A5;
    // Otherwise non-analog pin so return out of range pin to exit early
    default: return PINS_COUNT;
  }
}
```

```
2
3
4
5
6
7
```
This result makes sense in this loop, but if a board has pins "2, 3, 4, 5, 6 and 7" marked as ADC, you would expect the function call to be `analogRead(7)` instead of `analogRead(5)`.

On the other hand, the following lookup table prioritising digital pin values would give the following confusing result:

```c
_analogChannelPin(uint32_t pin) {
  switch (pin) {
    case 0:  return PIN_A0;
    case 1:  return PIN_A1;
    case PIN_A0:  return PIN_A0;
    case PIN_A1:  return PIN_A1;
    case PIN_A2:  return PIN_A2;
    case PIN_A3:  return PIN_A3;
    case PIN_A4:  return PIN_A4;
    case PIN_A5:  return PIN_A5;
    // Otherwise non-analog pin so return out of range pin to exit early
    default: return PINS_COUNT;
  }
}
``` 

```
2
3
2
3
4
5
```

sandeepmistry:
Hmm, maybe we're thinking too much about this. Some nRF5x boards are not really intended for Arduino users (like they don't have label Ax pins in the right spot or sequential). What if we just add this functionality for the ones that are based on boards like the Uno etc, where the Ax pins are all sequential?

carlosperate:
I think you are right there, if a board has both "pin numbering" and "Ax" numbers, then the manufacturer is most likely following the Arduino pin conventions. In that case the analogRead() function should accept both types of inputs. Otherwise, if the board doesn't have Ax pin number then the analogRead() function should consider its input to be only in the "generic pin number" format.

sandeepmistry:
@carlosperate ok, so back to something like https://github.com/sandeepmistry/arduino-nRF5/issues/133#issuecomment-293104322 then?

sandeepmistry:
Closing this for now due to lack of activity. Please let us know if/when this is ready for review and testing.

PR #152, opened 2017-04-24
Set nRF51 default ADC to 1/3 prescaler and 1/3 reference voltage

carlosperate:
So that the ADC input range is configured from 0 to Vdd instead of the current 0 to 3.6V, as discussed in #142.

### DIFF
--- a/cores/nRF5/wiring_analog_nRF51.c
+++ b/cores/nRF5/wiring_analog_nRF51.c
@@ -47,7 +47,7 @@ static struct PWMContext pwmContext[PWM_COUNT] = {
 
 static int timerEnabled = 0;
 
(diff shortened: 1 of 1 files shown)


carlosperate:
Probably a good idea as well. It looks like the internal reference is 0.6v so no easy way to do 0 to 3.3V with that.
Setting Vdd/4 as reference and keeping a gain of 4 would do the trick.
I could add that to the PR, but I don't really have an nRF52 board to test it.

sandeepmistry:
I'll merge this for now, we'll have to tackle the nRF52 side later

sandeepmistry:
@carlosperate thanks for the contribution!

carlosperate:
Sounds good, thanks!

PR #177, opened 2017-07-15
Add support for second SPI HW device

rogerclarkmelbourne:
As the nRF5x MUC's have hardware for 2 SPI devices, this change adds a second SPI device "SPI1" if SPI_INTERFACES_COUNT > 1 , as defined in variant.h

e.g.

```
/*
 * SPI Interfaces
 */
#define SPI_INTERFACES_COUNT 2

#define PIN_SPI_MISO         (27u)
#define PIN_SPI_MOSI         (31u)
#define PIN_SPI_SCK          (30u)

static const uint8_t SS   = 5u ;
static const uint8_t MOSI = PIN_SPI_MOSI ;
static const uint8_t MISO = PIN_SPI_MISO ;
static const uint8_t SCK  = PIN_SPI_SCK ;

#define PIN_SPI1_MISO         (4u)
#define PIN_SPI1_MOSI         (3u)
#define PIN_SPI1_SCK          (5u)

static const uint8_t SS1   = 7u ;
static const uint8_t MOSI1 = PIN_SPI1_MOSI ;
static const uint8_t MISO1 = PIN_SPI1_MISO ;
static const uint8_t SCK1  = PIN_SPI1_SCK ; 
```


sandeepmistry:
@rogerclarkmelbourne this is a bit tricky, from my understanding if the 2nd SPI is used then the I2C port can't be used.

Maybe we can live with it? cc/ @jacobrosenthal @carlosperate @dlabun

dlabun:
It's not the worst trade off in the world... The design I am working on right now with the Zero is all SPI devices.

sandeepmistry:
Ok, let's go for it them. A similar change could be done for I2C later as well.

sandeepmistry:
@rogerclarkmelbourne thanks for submitting!

PR #184, opened 2017-08-11
Fix stopBit = false mode in nRF51 Wire library

dsanders11:
The changes in #137 don't work with trying to use `stopBit = false`. Minor change to allow doing a `requestFrom` in multiple parts. This is necessary due to limited buffer sizes.

Example usage of what this allows (which doesn't currently work):

```
uint8_t buffer[20];

Wire.requestFrom(address, 10, false);
for (uint8_t i=0; i < 10; i++) {
    buffer[i] = Wire.read();
}

// Do something else, then finish reading

Wire.requestFrom(address, 10);
for (uint8_t i=0; i < 10; i++) {
    buffer[10 + i] = Wire.read();
}
```

@carlosperate, any thoughts on this PR? It's tested and works, but it's kind of quick and dirty as I needed it working ASAP.

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -84,6 +84,7 @@ class TwoWire : public Stream
     bool master;
     bool receiving;
     bool transmissionBegun;
+    bool suspended;
 
     // RX Buffer
     RingBuffer rxBuffer;
--- a/libraries/Wire/Wire_nRF51.cpp
+++ b/libraries/Wire/Wire_nRF51.cpp
@@ -36,6 +36,7 @@ TwoWire::TwoWire(NRF_TWI_Type * p_twi, uint8_t pinSDA, uint8_t pinSCL)
   this->_uc_pinSDA = g_ADigitalPinMap[pinSDA];
   this->_uc_pinSCL = g_ADigitalPinMap[pinSCL];
   this->transmissionBegun = false;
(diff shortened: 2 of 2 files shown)


dlabun:
@sandeepmistry Do you think this is a serious enough issue to include in v0.4.0?

sandeepmistry:
@dlabun yes, I think so.

@dsanders11 thanks for submitting! Do we need a similar change for `endTransmission(bool stopBit)`?
Apart from that, your changes look good to me (although I haven't tried them on a physical board yet).

dsanders11:
@sandeepmistry, probably needed in `endTransmission(bool stopBit)` as well, yea. I'm testing this on the [nRF52832 Breakout Board from Sparkfun](https://www.sparkfun.com/products/13990), and using `Wire_nRF51.cpp` despite it being nRF52 since that file was most recently improved and fixed, and works just fine since that API is available on nRF52. These changes could be applied to `Wire_nRF52.cpp` as well, but that file is still using the interrupt version which `Wire_nRF51.cpp` switched off of, so it may not cleanly apply.

sandeepmistry:
@dsanders11 thanks!

It would be awesome if you could spend some time on seeing what's wrong with the nRF52 Wire file. Maybe a combination of this PR and #137 need to be applied?

PR #194, opened 2017-09-28
Added support for the Arduino Primo Core

jeremypoulter:
Added support for the Arduino Primo Core. Basically grabbed Arduino.org's pin map from https://github.com/arduino-org/arduino-core-nrf52.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -252,6 +252,37 @@ STCT_nRF52_minidev.menu.softdevice.s132.build.extra_flags=-DNRF52 -DS132 -DNRF51
 STCT_nRF52_minidev.menu.softdevice.s132.build.ldscript=armgcc_s132_nrf52832_xxaa.ld
 
 
+Arduino_Primo_Core.name=Arduino Primo Core
+
+Arduino_Primo_Core.upload.tool=sandeepmistry:openocd
+Arduino_Primo_Core.upload.target=nrf52
+Arduino_Primo_Core.upload.maximum_size=524288
+
+Arduino_Primo_Core.bootloader.tool=sandeepmistry:openocd
+
+Arduino_Primo_Core.build.mcu=cortex-m4
+Arduino_Primo_Core.build.f_cpu=16000000
+Arduino_Primo_Core.build.board=PRIMO_CORE
+Arduino_Primo_Core.build.core=nRF5
+Arduino_Primo_Core.build.variant=PrimoCore
+Arduino_Primo_Core.build.variant_system_lib=
+Arduino_Primo_Core.build.extra_flags=-DNRF52
+Arduino_Primo_Core.build.float_flags=-mfloat-abi=hard -mfpu=fpv4-sp-d16
+Arduino_Primo_Core.build.ldscript=nrf52_xxaa.ld
+
+Arduino_Primo_Core.menu.softdevice.none=None
+Arduino_Primo_Core.menu.softdevice.none.softdevice=none
+Arduino_Primo_Core.menu.softdevice.none.softdeviceversion=
+
+Arduino_Primo_Core.menu.softdevice.s132=S132
+Arduino_Primo_Core.menu.softdevice.s132.softdevice=s132
+Arduino_Primo_Core.menu.softdevice.s132.softdeviceversion=2.0.1
+Arduino_Primo_Core.menu.softdevice.s132.upload.maximum_size=409600
+Arduino_Primo_Core.menu.softdevice.s132.build.extra_flags=-DNRF52 -DS132 -DNRF51_S132
+Arduino_Primo_Core.menu.softdevice.s132.build.ldscript=armgcc_s132_nrf52832_xxaa.ld
+
+
+
 # nRF51 variants
 ###################
 
--- a/variants/PrimoCore/variant.cpp
+++ b/variants/PrimoCore/variant.cpp
@@ -1,0 +1,144 @@
+/*
+  Copyright (c) 2016 Arduino Srl.  All right reserved.
+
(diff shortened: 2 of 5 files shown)


dlabun:
Before this can be merged the boards.txt and Readme also need to be updated. If you can update them that would be great, otherwise this will have to wait until somebody else is able to.

jeremypoulter:
Sure can do this, however with regard to the boards.txt, I added this primarily to get Primo Core support in Platform IO. For use with Arduino IDE would you not want to use Arduino.orgs core in preference? Adding to boards.txt could add confusion?

variants/PrimoCore/variant.cpp
```diff
@@ -0,0 +1,144 @@
...
+/*
+ * Pins descriptions
+ */
+//const PinDescription g_APinDescription[]=
+const uint32_t g_ADigitalPinMap[] = 
+{
+  // 0 .. 9 - Digital pins
+  // ----------------------
+  // 0..9
+  31, // { PORT0,  31, PIO_DIGITAL, (PIN_ATTR_DIGITAL|PIN_ATTR_PWM), ADC_A5, PWM0, NOT_ON_TIMER},
```
sandeepmistry: Can we remove the unneeded comments in this file?

variants/PrimoCore/variant.cpp
```diff
@@ -0,0 +1,144 @@
...
+  18, // { PORT0, 18, PIO_DIGITAL, PIN_ATTR_DIGITAL, No_ADC_Channel, NOT_ON_PWM, NOT_ON_TIMER}, // TP5
+  
+  //27..28 - NFC
+  // --------------------
+  9, // { PORT0, 9, PIO_DIGITAL, PIN_ATTR_DIGITAL, No_ADC_Channel, NOT_ON_PWM, NOT_ON_TIMER}, // NFC1
+  10, // { PORT0, 10, PIO_DIGITAL, PIN_ATTR_DIGITAL, No_ADC_Channel, NOT_ON_PWM, NOT_ON_TIMER} // NFC2
+
+} ;
+
+//Uart Serial(14, 15);
```
sandeepmistry: Why is the UART not declared?

variants/PrimoCore/variant.h
```diff
@@ -0,0 +1,143 @@
...
+
+/*----------------------------------------------------------------------------
+ *        Pins
+ *----------------------------------------------------------------------------*/
+
+// Number of pins defined in PinDescription array
+#define PINS_COUNT           (16u)
+#define NUM_DIGITAL_PINS     (8u)
+
+//#define digitalPinToPort(P)        ( NRF_P0 )
```
sandeepmistry: Do these comments need to be here?

sandeepmistry:
> Sure can do this, however with regard to the boards.txt, I added this primarily to get Primo Core support in Platform IO. 

I'm not familiar with Platform IO, how can the variant folder files be used without a boards.txt entry?

> For use with Arduino IDE would you not want to use Arduino.orgs core in preference? Adding to boards.txt could add confusion?

I'm confused, why is this PR is open if this is a concern then?

boards.txt
```diff
@@ -252,6 +252,37 @@ STCT_nRF52_minidev.menu.softdevice.s132.build.extra_flags=-DNRF52 -DS132 -DNRF51
...
 
 
+Arduino_Primo_Core.name=Arduino Primo Core
+
+Arduino_Primo_Core.upload.tool=sandeepmistry:openocd
+Arduino_Primo_Core.upload.target=nrf52
+Arduino_Primo_Core.upload.maximum_size=524288
+
+Arduino_Primo_Core.bootloader.tool=sandeepmistry:openocd
+
```
sandeepmistry: Does the Primo Core have a built-in programmer? (CMSIS DAP, JLINK, ST-LINK, etc?)

if so, a line should be added:
```
Arduino_Primo_Core.upload.protocol=<blah>
```

variants/PrimoCore/variant.cpp
jeremypoulter: Sure, just a hangover from Arduino's variant. Will remove all the commented out code.

variants/PrimoCore/variant.h
jeremypoulter: as above

variants/PrimoCore/variant.cpp
jeremypoulter: Not needed for this Core, Arduino's is structured slightly different

boards.txt
jeremypoulter: The Primo Core needs an external programmer

jeremypoulter:
Platform IO uses its own version of the build.txt file, so not needed. The reason I added this was two fold, it is not immediately obvious (in Platform IO) how to use multiple Arduino cores for the same chipset framework and they have already integrated this core for the Arduino/nRF52, and second is the Primo Core was a small stepping stone in our product development, both the prototype and subsequent versions will be based on generic nRF52 and use this core so I wanted to use this for the Primo Core version as well.

dlabun:
@sandeepmistry this one is bugging me... Do we really want to start supporting Arduino brand boards?

sandeepmistry:
@dlabun good point, let's leave this open for now and see what the demand is for it.

ttn115:
Hi, I have try the guide of README.md, but I was not able to flash my arduino primo core, 
anymore detail or advice?
https://github.com/sandeepmistry/arduino-nRF5/issues/217

dlabun:
Closing due to apparent lack of interest in supporting the Primo.

Issue #169, opened 2017-06-19
PWM signal from analogWrite does not stop when pin is set to different mode


carlosperate:
Looks the` analogWrite()` "pwm" signal on the nRF51 (now PWM hardware, so uses a timer instead) does not stop operation even after `pinMode()` and `digitalWrite()` has been applied to the same pin.

Simple sketch for testing:

```arduino
const int analogOutPin = 9;

void setup() {
  // put your setup code here, to run once:
  
}

void loop() {
  // Increase PWM until maximun value, takes around a couple of seconds
  for (byte i = 0; i<255; i++) {
    analogWrite(analogOutPin, i);
    delay(10);
  }

  // Toggle the same pin high and low a couple of times
  delay(1000);
  pinMode(analogOutPin, OUTPUT);
  digitalWrite(analogOutPin, HIGH);
  delay(1000);
  digitalWrite(analogOutPin, LOW);
  delay(1000);
  digitalWrite(analogOutPin, HIGH);
  delay(1000);
  digitalWrite(analogOutPin, LOW);
  delay(1000);
}
```

Expected behaviour:

* PWM signal increasing from minimum to maximum
* Same pin signal toggled high and low a couple of times
* loop iterates

Recorded behaviour:

* PWM signal increasing from minimum to maximum
* PWM signal stays at maximum value
* loop iterates

sandeepmistry:
@carlosperate thanks for reporting! Any suggestions on fixing this?

micooke:
The arduino core checks for this in digitalWrite. 
This adds a small overhead, but its probably best to be consistent.
 
Looking at the core it would be easy to fix (maybe as little as 3 lines), i can have a look into it if you wish.

sandeepmistry:
@micooke that would be great!

micooke:
Hmm, its turning out to be a bit trickier than i expected. I am working on it though. I need to  move a few static globals around so that they are accessible by wiring_digital.h. Ive implemented what i think should work, but it just doesnt at the moment. Ill keep you posted

micooke:
I think i fixed it locally, i will need someone to check on a nrf52832  though.
Linkage is a little 'special', so timer1 will stay on (not a biggie). 
I also added a fallback i found in the avr core that will use digitalWrite if all pwm channels are exhausted.
@sandeepmistry - it looks like pwm channel 0 is unavailable. Do you know if this is used by another peripheral?

micooke:
[...]

PR #193, opened 2017-09-17
Bugfix169 analog write

micooke:
* Fix bug #169
* wiring_analog_* : fallback to digitalWrite if no available PWM channel (copied from AVR core)
* wiring_analog_nRF52.c : convert pwmChannelPins & pwmChannelSequence -> pwmContext to semi-standardise pwm pin allocation and pwm status between nRF51 and nRF52
* wiring_private.h : Move pwm structures defines out of wiring_analog_* into wiring_private and make the instantiation of these structure externs instead of statics
* wiring_digital.c : disable the appropriate pwm timer when a digitalWrite is sent to an allocated (from analogWrite) pwm pin, and free up the pwm channel for re-allocation

PR #195, opened 2017-09-29
Fix bug #169

micooke:
wiring_analog_* : fallback to digitalWrite if no available PWM channel (copied from AVR core)
wiring_analog_nRF52.c : convert pwmChannelPins & pwmChannelSequence -> pwmContext to semi-standardise pwm pin allocation and pwm status between nRF51 and nRF52
wiring_private.h : Move pwm structures defines out of wiring_analog_* into wiring_private and make the instantiation of these structure externs instead of statics
wiring_digital.c : disable the appropriate pwm timer when a digitalWrite is sent to an allocated (from analogWrite) pwm pin, and free up the pwm channel for re-allocation

### DIFF
--- a/cores/nRF5/wiring_analog_nRF51.c
+++ b/cores/nRF5/wiring_analog_nRF51.c
@@ -28,27 +28,16 @@
 extern "C" {
 #endif
 
-#define PWM_COUNT 3
-#define PIN_FREE 0xffffffff
+static uint32_t adcReference = ADC_CONFIG_REFSEL_SupplyOneThirdPrescaling;
+static uint32_t adcPrescaling = ADC_CONFIG_INPSEL_AnalogInputOneThirdPrescaling;
 
-struct PWMContext {
-  uint32_t pin;
-  uint32_t value;
-  uint32_t channel;
-  uint32_t mask;
-  uint32_t event;
-};
-
-static struct PWMContext pwmContext[PWM_COUNT] = {
+struct PWMContext pwmContext[PWM_COUNT] = {
   { PIN_FREE, 0, 1, TIMER_INTENSET_COMPARE1_Msk, 1 },
   { PIN_FREE, 0, 2, TIMER_INTENSET_COMPARE2_Msk, 2 },
   { PIN_FREE, 0, 3, TIMER_INTENSET_COMPARE3_Msk, 3 }
 };
 
-static int timerEnabled = 0;
-
-static uint32_t adcReference = ADC_CONFIG_REFSEL_SupplyOneThirdPrescaling;
-static uint32_t adcPrescaling = ADC_CONFIG_INPSEL_AnalogInputOneThirdPrescaling;
+struct PWMStatus pwmStatus[PWM_TIMER_COUNT] = {0, TIMER1_IRQn};
 
 static uint32_t readResolution = 10;
 static uint32_t writeResolution = 8;
@@ -229,7 +218,7 @@ void analogWrite( uint32_t ulPin, uint32_t ulValue )
 
   ulPin = g_ADigitalPinMap[ulPin];
 
-  if (!timerEnabled) {
+  if (pwmStatus[0].numActive == 0) {
     NVIC_SetPriority(TIMER1_IRQn, 3);
     NVIC_ClearPendingIRQ(TIMER1_IRQn);
     NVIC_EnableIRQ(TIMER1_IRQn);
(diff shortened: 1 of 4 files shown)


micooke:
Tidy-up of #193

cores/nRF5/wiring_analog_nRF51.c
```diff
@@ -259,17 +246,25 @@ void analogWrite( uint32_t ulPin, uint32_t ulValue )
...
       NRF_TIMER1->INTENSET = pwmContext[i].mask;
-
-      break;
+      pwmStatus[0].numActive++;
+      return;
     }
   }
+
+  // fallback to digitalWrite if no available PWM channel
+  if (ulValue < 128)
```
sandeepmistry: Can the write resolution be higher than 8-bits? If so, we need to adjust this.

cores/nRF5/wiring_digital.c
```diff
@@ -86,18 +93,42 @@ void digitalWrite( uint32_t ulPin, uint32_t ulVal )
 
   ulPin = g_ADigitalPinMap[ulPin];
 
+  for (uint8_t i = 0; i < PWM_COUNT; i++)
```
sandeepmistry: What do you think about moving this to a private `analogWriteOff` (or something with a better name) function in `cores/nRF5/wiring_analog_nRF5x.c`?

micooke: Rather than creating another function I could change analogWrite so that it turns off PWM when a duty cycle of zero is passed?

Its up to what you prefer as you are the owner and this affects the API. Either implementation is easy to implement.

cores/nRF5/wiring_analog_nRF51.c
micooke: On it. One thing i noticed is that the read and write resolution isnt bounded and for the NRF52 they are stored as signed integers so you could set the read or write resolution to +/-2e9!

If its alright ill add a static int with the half max value and a switch statement to set it in writeResolution so that valid values are used (and so they arent re-calculated on each analogWrite)

micooke:
Ok, can of worms officially opened :boom:
nrf51 is done, but i noticed something i should have noticed earlier - the nRF52 only uses 3 out of 12 PWM channels. It has 3 PWM/TIMERS with 4 compare channels each, and in the core each compare channel output pin points to the same pin - so 3 outputs max (currently). This will need some work to fix it up, and luckily now i have the hardware to test it on!

micooke:
Sorry i should have asked, @sandeepmistry , @dlabun - do you want me to fix the nRF52 PWM channel issue here or in a separate PR?

dlabun:
Sure, go for it.

cores/nRF5/wiring_analog_nRF51.c
```diff
@@ -259,17 +246,25 @@ void analogWrite( uint32_t ulPin, uint32_t ulValue )
...
+  {
+    digitalWrite(ulPin, LOW);
+  }
+  else
+  {
+    digitalWrite(ulPin, LOW);
```
micooke: Should be HIGH - copy/paste error

cores/nRF5/wiring_analog_nRF52.c
```diff
@@ -236,15 +235,25 @@ void analogWrite( uint32_t ulPin, uint32_t ulValue )
...
   }
+
+  // fallback to digitalWrite if no available PWM channel
+  if (ulValue < 128)
+  {
+    digitalWrite(ulPin, LOW);
+  }
+  else
+  {
+    digitalWrite(ulPin, LOW);
```
micooke: The same copy/paste error again

cores/nRF5/wiring_analog_nRF51.c
micooke: For reference;
* nRF51 TIMER1 (the only timer currently used for PWM) is either 8b or 16b write resolution
* nRF52 is up to 32b write resolution for all 12 PWM channels

micooke:
Update : 
* 6x analog/PWM signals enabled for nRF51 (up from 3)
* 12x analog/PWM signals enabled for nRF52 (up from 3)

The nRF5x documentation is lacking pretty severely compared to the AVR chips I have dealt with. Im **very** surprised but the implementation for the nRF51 uses software to set/clear the hardware pins. It also uses the first channel as the period interrupt, rather than having a separate overflow interrupt - as per RBL and several Nordic source code.

It basically looks like PWM is implemented rather poorly in the nRF51, and i guess that is the reason that the nRF52 has dedicated PWM channels.

dlabun:
@micooke Any issues with merging this fix?

sandeepmistry:
Let’s hold off on this one for a bit ...

micooke:
Hi @sandeepmistry, is there an issue with this PR I haven't addressed or does it just need more testing?

sandeepmistry:
@micooke @dlabun this one needs some more testing.

I'm concerned of the performance implications of calling `analogWrite(pin, 0)` for every `digitalWrite(pin, value)` call.

micooke:
That bit was actually copied from the Arduino AVR core

Issue #199, opened 2017-10-03
I2C freeze. Custom NRF52


snowneki4:
I have a custom nrf52 board and i need to get MPU6050 running. I have connected it using following pins:
#define PIN_WIRE_SDA         (3u)
#define PIN_WIRE_SCL         (21u)
Both pins have 4.7k resistors connected.
I have not yet implemented MPU6050 library, though i am trying to recognize the device using I2C scanner described here: [https://playground.arduino.cc/Main/I2cScanner](url) Code i am using is there.
Whenever sensor is disconnected terminal continues to show No I2C devices found. Which is right. But whenever i connect sensor to the pins it freezes after the scanning stage. 
So, questions. Originally pins for SDA/SCL were 20/21 in variants.h. Can i actually change those? 
Can you point me what am i doing wrong? 

My skills arent very high, so could be i am doing something completely stupid.

micooke:
See #176

lichtheini:
Maybe I can help:  I have an Adafruit Feather nRF52 and a MPU9250, so much the same condition. There are many points were your code or wiring could be wrong. I just show the one, that took me a while to figure out: a hanging I2C-bus with SDA low. 

My solution, well, my workaround is described  here: [https://github.com/esp8266/Arduino/issues/1025#issuecomment-319199959](https://github.com/esp8266/Arduino/issues/1025#issuecomment-319199959) and [here](http://www.forward.com.au/pfod/ArduinoProgramming/I2C_ClearBus/index.html).

```
if(digitalRead(PIN_WIRE_SCL) == HIGH && digitalRead(PIN_WIRE_SDA) == LOW) {
      //Serial.println("reset i2c bus");
      pinMode(28, OUTPUT);  // connect parallel to SCL
      digitalWrite(28, LOW);
      delay(200); 
      pinMode(28, INPUT);
      delay(50);
      NVIC_SystemReset();
}
```

And yes, almost any pins are usable for I2C communication. 

> The GPIOs used for each two-wire interface line can be chosen from any GPIO on the device and are
independently configurable.  (Product Spec  rev 1.3, p. 305)

Hope it helps!

snowneki4:
Problem was somehow solved my selecting Generic NRF51 option from the board selection menu, even
though i am using nrf52 board.
It is described in the post micooke has mentioned.
thanks everyone

micooke:
Glad it worked for you!
[...]

Issue #108, opened 2017-03-02
Wire library


kriswiner:
I am using my own custom nRF52832 development boards and I would like to initialize the I2C bus, which I have placed on pins 6 (SDA) and 7 (SCL) out of convenience. I looked in the nRF52 DK variant files and tried to change the pin numbers there but it is pretty confusing.  I managed to figure out the pin labelling for the leds of the board but how can I enable I2C on pins 6 and 7. Is the default on pins 26 and 27 or pins 20 and 21, and are these pins the P0.20 and P0.21 pins or physical pins 20, 21 or what?

Wouldn't it make more sense to have a general nRF52 variant and let the user choose as in:

Serial.begin(RX_pin, TX_pin, baudrate);  and

Wire.begin(SDA_pin, SCL_pin); ?

Also, I tried to use TimerOne.h from the BLEPeripheral library examples and the IDE complains that Timer1.initialize and Timer1.attachinterrupt are not a Timer1 class.

Is there no other way to use timers for the nRF52?

Thank you for our help!

kriswiner:
I was able to change the SDA/SCL pins in the variant.cpp to the correxct pins 6 and 5 in my case but I cannot get an ACK on the I2C Scanner. The SCL line is being held down tp 0.11 V. Not quite sure I understand this behavior. I tried on another pair of pins and I see the same behavior. Something in the Wire implementation is holding the SCL line. 

Is there any example of writing to a sensor and reading its data?

kriswiner:
I am getting error = 2 or NACK on all addresses up to 0x67 and then on 0x68, which is the address of the MPU9250, an I2C device on the bus, I am getting nothing from endTransmission, the bus is hanging. Do you know what the reason is for this?

dlabun:
What value are your I2C pull up resistors?

kriswiner:
4K7, this board works fone with mbed, it is not a hardware issue.

dlabun:
What version of the core are you using?

kriswiner:
0.2.0

Is there a newer or better one?

dlabun:
v0.3 is out now, and we did make a change to the configuration of the I2C bus. You may want to try it and see if there is any difference.

kriswiner:
updating now...

dlabun:
Have you tried any of the libraries for the MPU9250 ? I can't judge if your scanner code is working correctly or not.

kriswiner:
[...]

PR #207, opened 2017-10-20
disable DMA I2C on BLE Nano 2

gtalusan:
This will at least get I2C working on the RedBear BLE Nano 2.

I don't have any different boards for which this is a problem, so I'm assuming Wire_nRF52 is working for everything else except the Nano 2.

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -31,16 +31,20 @@
 // WIRE_HAS_END means Wire has end()
 #define WIRE_HAS_END 1
 
+#if defined(NRF52) && !defined(ARDUINO_RB_BLE_NANO_2)
+#define USE_DMA_TWI 1
+#endif
+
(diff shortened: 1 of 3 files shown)


dlabun:
This is the first I am hearing about a problem with the BLE Nano 2 having I2C issues, are you sure it's related to the I2C library?

gtalusan:
Whoops, my bad.  That's what I get for reading the world news in another tab.  Updated the commit.

#108 and #199 talk about problems with I2C on the nRF52.  At the moment I'd like to just get I2C working with this core.

I2C works with the RedBearLabs core, but it doesn't use DMA either.

dlabun:
Since there's a generic problem with I2C on the nRF52 the fix should be generic to all of the nRF52 boards and not just target a single board.

gtalusan:
You're right.  No worries, I'll just leave this in my fork now.

Issue #206, opened 2017-10-17
nRF52 - use NFC pins as GPIO


micooke:
I know this adds yet another compile and variant option, but for nRF52 boards PO.09 and PO.10 are by default reserved for a NFC antenna. Are we able to support an option under variants to enable/disable this option and hence free up two additional pins if you aren't using NFC?

To use these pins as GPIO you need to #define CONFIG_NFCT_PINS_AS_GPIOS

micooke:
To add more to this, as mentioned in a previous issue (that i cant find) it would be nice to add an option to enable/disable PO.29 as a reset pin via #define CONFIG_GPIO_AS_PINRESET

dlabun:
For the reset, check PR #91 as there's already some discussion there.

micooke:
Maybe we can bundle them into one variant selection field with four options? They are both GPIO configuration options.

dlabun:
Not a bad idea at all

micooke:
I have this working locally, i csn spawn a PR on Monday with an example then once you are happy with the variant wording i can propogate it to the other boards.

dlabun:
Awesome. I am also thinking at this point we should probably start a changelog.

micooke:
Sounds like a plan Stan.

There have been quite a few bug fixes & enhancements since the last release so good to start capturing this. 

Im also looking in to #176 - the arduino core lib is quite different, i want to see if that will resolve the issue.

dlabun:
Something to keep in mind, we are using an older version of Nordic SDK... Maybe we're bumping up against limitations or bugs in the older SDK.

netkruzer:
I put "#define CONFIG_NFCT_PINS_AS_GPIOS" in my INO file an I can't seem to get pins p0.09 and p0.10 to work as inputs.

I'm using Generic nRF52.

micooke:
The SDK gets compiled separately and linked in, so you need that ```#define``` at compile time. If you check out PR #208, there is a board configuration option that does this for the generic NRF52

netkruzer:
I know this might be a dumb question, but how do I go about activating this on my build? Do I modify the boards.txt file or something else?

micooke:
Yep, you could just append ```-DCONFIG_NFCT_PINS_AS_GPIOS``` to build.extra_flags on line 42.

Note that if you are using the soft device, you will need to append it to line 54.

Or you could just pull the changes i made in #208 :wink:

[...]

PR #91, opened 2017-02-09
Support standard reset button as default.

bojh:
Nordic does provide a standard mapping of a reset functionality on GPIO pin P.21 which is in most boards the reset button. To support the RESET button in a standard Arduino way, we have to provide a C compile flag CONFIG_GPIO_AS_PINRESET.

PR #208, opened 2017-10-23
Enhancement : Add GPIO pin mapping functions

micooke:
GPIO pin mapping options added to platform.txt and GPIO pin mapping options added example Generic_nRF52832 example:
* default=No Pin Reset, NFC pins as Antenna (default)
* PinResetNFCAntenna=Pin Reset, NFC pins as Antenna
* NoPinResetNFCGPIO=No Pin Reset, NFC pins as GPIO
* PinResetNFCGPIO=Pin Reset, NFC pins as GPIO

In support of issue #206 : nRF52 - use NFC pins as GPIO and as an alternative to PR #91 : Support standard reset button as default.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -20,6 +20,7 @@ menu.softdevice=Softdevice
 menu.version=Version
 menu.lfclk=Low Frequency Clock
 menu.board_variant=Board Variant
+menu.gpio_config=GPIO Configuration
 
 # nRF52832 variants
 ###################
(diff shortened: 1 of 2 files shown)


micooke:
Let me know what you think of the wording in boards.txt and ill replicate it for the other nrf52 boards.

sandeepmistry:
@micooke thanks for submitting!

It feels like the generic variant is getting a bit complex with all these options. I'm leaning towards making them strictly targeted to a non-generic variant at this point. @dlabun what do you think?

micooke:
To explain, not everyone uses NFC and not everyone will want a reset button. They may prefer to free it up as GPIO for their application. Personally i find the reset pin only useful during debug, then i normally want to free it up.

Hence the options.

This is obviously covers some of the same ground as #212, but relies on the sdk to setup the reset pin.

sandeepmistry:
Understood, but for these more complex scenarios, maybe advanced users can create their own variant?

micooke:
Does anyone else want this? Otherwise I'm happy to drop this, and just support it on my smartwatch repo?

sandeepmistry:
@micooke I'd say to put it in the smart watch repo for now.

Martin21-hw:
Hello,

Although this thread may be a bit dated, I was eager to try implementing the suggested changes by modifying the Boards.txt and Platform.txt files. Despite successfully compiling the code, I unfortunately did not see any additional menu items. I am curious if anyone has insights as to why this may be occurring.

Thank you in advance for any assistance or suggestions you may have.

PR #212, opened 2017-11-20
set a reset pin if RESET_PIN is defined in variant.h

ntavish:
This seems to be the minimal change that can be used by variants to set a RESET pin. The reset pin persists till a full erase.


micooke:
That's a neat way to do it!

cores/nRF5/main.cpp
```diff
@@ -16,10 +16,30 @@
...
+      NRF_UICR->PSELRESET[0] = RESET_PIN;
+      while (NRF_NVMC->READY == NVMC_READY_READY_Busy){}
+      NRF_UICR->PSELRESET[1] = RESET_PIN;
+      while (NRF_NVMC->READY == NVMC_READY_READY_Busy){}
+      NRF_NVMC->CONFIG = NVMC_CONFIG_WEN_Ren << NVMC_CONFIG_WEN_Pos;
+      while (NRF_NVMC->READY == NVMC_READY_READY_Busy){}
+      NVIC_SystemReset();
+  }
+  #endif
+}
```
sandeepmistry: I think this is better suited to go in [wiring.c](https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/wiring.c). In case a variant overrides `initVariant`.

sandeepmistry:
@ntavish thanks!

PR #214, opened 2017-12-06
Fix I2C lock-up when endTransmission() is called with empty txBuffer

alextsam:
This is fixing the lock-ups that are described in #108, #207, and #199. The problem (at least in the case of the I2C scanner that I had to deal with) is related to an empty tx buffer, so the LASTTX event never occurres, resulting into an endless loop.

### DIFF
--- a/libraries/Wire/Wire_nRF52.cpp
+++ b/libraries/Wire/Wire_nRF52.cpp
@@ -214,7 +214,9 @@ uint8_t TwoWire::endTransmission(bool stopBit)
   while(!_p_twim->EVENTS_TXSTARTED && !_p_twim->EVENTS_ERROR);
   _p_twim->EVENTS_TXSTARTED = 0x0UL;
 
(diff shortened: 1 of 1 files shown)


alextsam:
Alternatively, maybe we should not set-up TX at all in endTransmission() when the buffer is empty. What do you think?

sandeepmistry:
@alextsam I think an `if` statement is better for the `0` send length case.

Does the I2C scanner work with this change? Maybe you can share the sketch you are using and output?

alextsam:
Sorry, I kind of forgot about this. I will change it to not setup the DMA TX in case of 0-length. I expect to do this tonight (CET)

alextsam:
@sandeepmistry @dlabun I changed the commit to have an if guard around the while instead as suggested. When I tested it (at the time of the original post) I tested with 2 I2C sensors without any issues whatsoever as well as with the default example I2C scanner without any lock-ups anymore

sandeepmistry:
@alextsam perfect, thanks for this!

PR #219, opened 2018-01-03
Added Sino:bit board

ScepticInt:
Hi I added support for Sino:bit by Naomi Wu and Elecrow (https://github.com/sinobitorg/)
I also added Dave Astels (https://github.com/dastels/Sinobit) Sinobit library and Adafruits HT1632 /GFX libraries. Since the Led Matrix on Sino:Bit is controlled by a HT1632 and not pins like Micro:Bit these files (IMHO) count as required for a complete "out of the box" experience.


dlabun:
Please remove the 3rd party libraries, we don't maintain them in this repository. You don't need to modify the JSON file, it gets modified just as we do a new release.

ScepticInt:
Done!

variants/Sinobit/variant.h
```diff
@@ -0,0 +1,104 @@
...
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Lesser General Public License for more details.
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#ifndef _VARIANT_BBC_MICROBIT_
```
sandeepmistry: This needs to be updated ...

boards.txt
```diff
@@ -818,3 +818,50 @@ ng_beacon.menu.softdevice.s130.softdeviceversion=2.0.1
...
+Sinobit.upload.setup_command=transport select swd;
+Sinobit.upload.use_1200bps_touch=false
+Sinobit.upload.wait_for_upload_port=false
+Sinobit.upload.native_usb=false
+
+Sinobit.bootloader.tool=sandeepmistry:openocd
+
+Sinobit.build.mcu=cortex-m0
+Sinobit.build.f_cpu=16000000
+Sinobit.build.board=BBC_MICROBIT
```
sandeepmistry: This should be updated as well.

boards.txt
```diff
@@ -818,3 +818,50 @@ ng_beacon.menu.softdevice.s130.softdeviceversion=2.0.1
 ng_beacon.menu.softdevice.s130.upload.maximum_size=151552
 ng_beacon.menu.softdevice.s130.build.extra_flags=-DNRF51 -DS130 -DNRF51_S130
 ng_beacon.menu.softdevice.s130.build.ldscript=armgcc_s130_nrf51822_xxaa.ld
+
+
+Sinobit.name=Sino:bit
+
+Sinobit.vid.0=0x0d28
```
sandeepmistry: Since we don't have a sino:bit to test with, are these the correct USB id's (same as micro:bit)?

variants/Sinobit/variant.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#include "variant.h"
+
+const uint32_t g_ADigitalPinMap[] = {
+  // 0 - 4
```
sandeepmistry: Are the comments in this file valid? If not please remove or update.

boards.txt
ScepticInt: Yes, It's the MBED, cmsis-DAP 0d28:0204

boards.txt
ScepticInt: Is that a compiler parameter or something i can set to any arbitrary string?

ScepticInt:
All done except, Sinobit.build.board=BBC_MICROBIT . waiting for answer.

boards.txt
sandeepmistry: Compiler parameter, would get mapped to `ARDUINO_<value>` as a compile time define (`-DARDUINO_<value>`).

ScepticInt:
Correction, I did push the Sinobit.build.board. I think all CR's are solved.

.travis.yml
```diff
@@ -35,3 +35,4 @@ script:
   - buildExampleSketch sandeepmistry:nRF5:TinyBLE 01.Basics Blink
   - buildExampleSketch sandeepmistry:nRF5:bluey 01.Basics Blink
   - buildExampleSketch sandeepmistry:nRF5:hackaBLE 01.Basics Blink
+  - buildExampleSketch sandeepmistry:nRF5:Sinobit 01.Basics BareMinimum
```
dlabun: Shouldn't this be Blink like all of the other boards?

ScepticInt:
There's no LED on gpio pin on rev 1. Only on LED driver which we'd need the ht1632 lib for.
So without extra source files we unfortunately can't blink.  I removed LED1 from pin config.

dlabun:
In that case, there's no need to list the Blink example since it's effectively useless.

ScepticInt:
Where are the examples listed? i cannot find the source for those in this repo.

dlabun:
I believe they're part of the Arduino IDE (I think)

sandeepmistry:
Yes, it's part of the IDE.

@dlabun let's go with using BareMinimum for the Travis CI test as @ScepticInt suggested for now. I'll merge this shortly ...

dlabun:
Works for me.

ScepticInt:
Thanks! Den 14 jan. 2018 17:53 skrev Sandeep Mistry <notifications@github.com>:Merged #219.

Issue #213, opened 2017-11-21
Custom I2C/SPI/UART pins for generic variants


honnet:
Hi again @sandeepmistry & @dlabun ;)

Cheap [nRF52 modules](https://www.aliexpress.com/item/wholesale-Holyiot-TinyBLE-nRF52832-Bluetooth-low-energy-module-BLE-5-0-for-Bluetooth-mesh/32840369737.html) are getting common, so this [Generic variant](https://github.com/sandeepmistry/arduino-nRF5/blob/master/variants/Generic/) that you offer here is really a great feature.

Currently, the the default Wire object is initialized with default I2C pins that are imposed:
- [Wire_nRF52.cpp#L387](https://github.com/sandeepmistry/arduino-nRF5/blob/master/libraries/Wire/Wire_nRF52.cpp#L387)
- [Generic/variant.h#L96-L100](https://github.com/sandeepmistry/arduino-nRF5/blob/master/variants/Generic/variant.h#L96-L100)

It makes sense for most boards, but maybe not for the generic variants as these pins could be anywhere.

It would be a pleasure to implement + PR an improvement allowing more flexibility but it seems appropriate to hear your thoughts about the following possible solutions:

1. Modifying a bit the declaration/implementation of `begin()` for [master & slave](https://github.com/sandeepmistry/arduino-nRF5/blob/master/libraries/Wire/Wire.h#L42-L44) to allow setting I2C pins with optional default values (as in [here](http://en.cppreference.com/w/cpp/language/default_arguments)) might be the most appropriate.
2. Using `#define` instead of `const` [here](https://github.com/sandeepmistry/arduino-nRF5/blob/master/variants/Generic/variant.h#L99-L100), should allow using `#undef` but it might not be very reliable, so simply removing the [default values](https://github.com/sandeepmistry/arduino-nRF5/blob/master/variants/Generic/variant.h#L70-L100) seems more convenient. It would also take care of [SPI](https://github.com/sandeepmistry/arduino-nRF5/blob/master/variants/Generic/variant.h#L80-L89) and [UART](https://github.com/sandeepmistry/arduino-nRF5/blob/master/variants/Generic/variant.h#L74-L75), and users would just have to declare/define the pin constants/macros in their personal files).

Any thought?
Cedric ;)

dlabun:
The generic device is really more of a template, if you're building a custom device you really should be taking the generic device and modifying it to your needs. Of course, that's not to say we can't make the process more robust.

[...]

PR #221, opened 2018-01-06
Custom I2C/SPI/UART pins for generic variants

honnet:
Hi again @sandeepmistry & @dlabun!

As said in the issue #213, this Generic variant that you offer here is a really great feature, but the default Wire/Serial/SPI objects are initialized with default pins that are forced.
It makes sense for most boards, but maybe not for the generic variants as these forced pins could be unavailable.

The 3 commits of this pull requests implement a setPins(...) method for the 3 objects aforementioned (Wire/Serial/SPI), calling it before its respective begin() would solve the problem.

Let me know if there's anything I can do to improve these commits...
Cedric ;)

### DIFF
--- a/cores/nRF5/Uart.cpp
+++ b/cores/nRF5/Uart.cpp
@@ -41,6 +41,22 @@ Uart::Uart(NRF_UART_Type *_nrfUart, IRQn_Type _IRQn, uint8_t _pinRX, uint8_t _pi
   uc_hwFlow = 1;
 }
 
+#ifdef ARDUINO_GENERIC
+void Uart::setPins(uint8_t _pinRX, uint8_t _pinTX)
+{
+  uc_pinRX = g_ADigitalPinMap[_pinRX];
+  uc_pinTX = g_ADigitalPinMap[_pinTX];
+}
+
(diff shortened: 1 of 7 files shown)


sandeepmistry:
@honnet thanks for submitting!

What do you think about making these calls enabled per variant? (so maybe Generic only for now), we can wrap them like `#ifdef NRF5_ENABLE_SET_PINS`, then the variant can have a `#define NRF5_ENABLE_SET_PINS`.

cc/ @dlabun

honnet:
Sounds good!
We could take advantage of the flag already used when compiling code for generic boards (see `-DARDUINO_{build.board}`), and it would work for nRF51 as well as nRF52:
https://github.com/sandeepmistry/arduino-nRF5/blob/master/platform.txt#L73-L76
https://github.com/sandeepmistry/arduino-nRF5/blob/master/boards.txt#L37
https://github.com/sandeepmistry/arduino-nRF5/blob/master/boards.txt#L305

So we would use `#ifdef ARDUINO_GENERIC`, is it OK for you?
Cedric ;)

sandeepmistry:
@honnet sounds good, thanks for the contribution :)

honnet:
Thanks to you!
It's always a pleasure ;)

PR #224, opened 2018-01-14
Add -D{board.variant} to build.extra_flags of PCA1000X softdevices

oyooyo:
The _build.extra_flags_ for PCA1000X boards include `-D{board.variant}`:
`PCA1000X.build.extra_flags=-DNRF51 -D{board.variant}`
However, this extra flag is currently not also part of the _build.extra_flags_ of the PCA1000X softdevices:
```
PCA1000X.menu.softdevice.s110.build.extra_flags=-DNRF51 -DS110 -DNRF51_S110
[...]
PCA1000X.menu.softdevice.s130.build.extra_flags=-DNRF51 -DS130 -DNRF51_S130
```
Because of this, `-D{board.variant}` will actually only be used on PCA1000X boards when selecting softdevice "none".
On PCA1000X boards that require hardware flow control for UART communication (like the PCA10000 and PCA10001), this for example leads to the strange effect that serial port communication currently only works when selecting softdevice "none".

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -465,14 +465,14 @@ PCA1000X.menu.softdevice.s110=S110
 PCA1000X.menu.softdevice.s110.softdevice=s110
 PCA1000X.menu.softdevice.s110.softdeviceversion=8.0.0
 PCA1000X.menu.softdevice.s110.upload.maximum_size=151552
-PCA1000X.menu.softdevice.s110.build.extra_flags=-DNRF51 -DS110 -DNRF51_S110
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@oyooyo these changes look ok to me, could you please also update the Travis CI file to leverage the new variant, see: https://github.com/sandeepmistry/arduino-nRF5/blob/master/.travis.yml#L30-L31

You'll just need a new line.

oyooyo:
I believe I was successful at removing the duplicate commit, so this PR should be okay now.

PR #229, opened 2018-01-16
define Reset-Pin for nRF52 platforms. (see#91)

bojh:
Signed-off-by: Hubschneider Joerg (INST-IOT/ESW-Wa) <joerg.hubschneider@bosch-si.com>

### DIFF
--- a/variants/Generic/variant.h
+++ b/variants/Generic/variant.h
@@ -99,11 +99,6 @@ static const uint8_t SCK  = PIN_SPI_SCK ;
 static const uint8_t SDA = PIN_WIRE_SDA;
 static const uint8_t SCL = PIN_WIRE_SCL;
 
-/*
- * Reset Button at P0.21
- */
-#define RESET_PIN            21
-
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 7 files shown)


variants/Generic/variant.h
```diff
@@ -99,6 +99,11 @@ static const uint8_t SCK  = PIN_SPI_SCK ;
 static const uint8_t SDA = PIN_WIRE_SDA;
 static const uint8_t SCL = PIN_WIRE_SCL;
 
+/*
+ * Reset Button at P0.21
+ */
+#define RESET_PIN            21
```
sandeepmistry: @bojh the Travis CI build is failing, for nRF51,

I suggest we either:

1) Remove it for the generic variant, people might want to use the pin as a GPIO

or

2) Wrap it in an `#ifdef NRF52` for it's only possible for nRF52 boards.

I prefer option 1) but could be convinced about 2) :)

@dlabun any feedback on this?

bojh: @sandeepmistry May option 1 is more common, So I will remove it for the generic variant.

PR #263, opened 2018-03-29
Update URL DFU S132

BurlakovSG:


dlabun:
Last I checked with Nordic the newer SD was not compatible with the ones included in this project.

dlabun:
Closing, no response from OP.

PR #311, opened 2018-09-10
Add Seeed Studio Arch Link support

wodu:
[[http://wiki.seeedstudio.com/Arch_Link/]]

Signed-off-by: Wojciech Dubowik <Wojciech.Dubowik@gmail.com>


sandeepmistry:
Hi @wodu,

Thanks for submitting this pull request!

Since this board has an Arduino Uno compatible connector, I think it would be better for the mapping in `variants/ArchLink/variant.cpp` to reflect this and map to the diagram on: http://wiki.seeedstudio.com/Arch_Link/

In this way `digitalWrite(0, HIGH)` would control the p7 pin in the diagram.

wodu:
Hello,
  Yes, I will change the pins and leave in comments pin name from the board labels. The latter was the reason why I have kept the original mapping.

Wojtek

wodu:
I have pushed requested changes.

Wojtek

jandrassy:
I have this board (2 days now). I agree with the Uno numbering. It is best for shields. I made my own mapping, the I discovered this PR.
SS should be 10 and SDCARD_SS  should be 4. They are remapped too with g_ADigitalPinMap.
There is a red LED on P13/D7. I have
#define PIN_LED1             (7)  // P0.13
#define PIN_LED2             (20) // P0.30
#define LED_BUILTIN          PIN_LED1
I am not sure about AREF as pin.
The P30 is supposedly A6 according to schematics in pdf.

jandrassy:
sorry, pin D7/P17 is connected to W5500 reset

jandrassy:
To support this board, I updated my [OTEthernet library](https://github.com/jandrassy/OTEthernet), which is based on WiFi101OTA library. I added nRF51 support to InternalStorage.cpp

boards.txt
```diff
@@ -948,3 +948,42 @@ DWM1001-DEV.menu.softdevice.s132.softdeviceversion=2.0.1
 DWM1001-DEV.menu.softdevice.s132.upload.maximum_size=409600
 DWM1001-DEV.menu.softdevice.s132.build.extra_flags=-DNRF52 -DS132 -DNRF51_S132
 DWM1001-DEV.menu.softdevice.s132.build.ldscript=armgcc_s132_nrf52832_xxaa.ld
+
+ArchLink.name=Seeed Arch Link
```
sandeepmistry: Let's going with @jandrassy suggestion from https://github.com/sandeepmistry/arduino-nRF5/pull/311#issuecomment-441438774 and change the key to SeeedArchLink in boards.txt

variants/ArchLink/variant.h
```diff
@@ -0,0 +1,105 @@
...
+/*
+ * SPI Interfaces
+ */
+#define SPI_INTERFACES_COUNT 1
+
+#define PIN_SPI_MISO         (12)
+#define PIN_SPI_MOSI         (11)
+#define PIN_SPI_SCK          (13)
+
+static const uint8_t SS   = 22 ;
```
sandeepmistry: Again from @jandrassy's comments (https://github.com/sandeepmistry/arduino-nRF5/pull/311#issuecomment-441438774), I think this should be 10 right?

jandrassy:
I am still in the process of adapting this board for my project. I learn the details of the board. Then if @wodu doesn't make the changes I will make a PR with my changes. The board.txt lines needs some attention too.

wodu:
I will test these changes and push the changes. I have one test which makes the board MySensors Ethernet gateway and try to test SD card again.

wodu:
I have merged and pushed the changes.

jandrassy:
I like the changes. Only nit-picking: in schematics of Arch Link, P30 is labeled A6 ADC. AVR has A6 as pin 20.

wodu:
I have added A6 definition. Well spotted.

jandrassy:
The W5500 reset and interrupt pins are non-standard for Arduino. The Arduino Ethernet library doesn't use them. The Ethernet shield has reset connected to board reset, but on this board the reset is handled by LPC11U35. **Perhaps a define for the W5500 reset pin would be useful.** The Ethernet library does only a software reset of W5500. I don't know now, if a hardware reset in setup() will be necessary.

Note: The interrupt of W5500 can't be used with Ethernet library and W5500 holds the pin HIGH with pull-up.

wodu:
From what I can read from code hw reset is not necessary. I have been using it for a while and I havn't noticed any issues with standard Ethernet examples.

jandrassy:
then OK. unfortunate is, that the pin 7 can't be used for anything else.

@sandeepmistry, please merge this PR before the release

jandrassy:
@wodu, I tested from Arduino IDE (I use Eclipse) and it didn't upload. this line was missing in board.txt

`SeeedArchLink.upload.setup_command=transport select swd;`

wodu:
It looks like it's setup problem. I use linux and for me openocd selects swd automatically. We might need to force it as I get following info
`Info : auto-selecting first available session transport "swd". To override use 'transport select <transport>'.`

If you are using windows, could you test my archlink_support branch. I had to change openocd command lines to work on linux but I don't know whether it breaks other platforms.

jandrassy:
I am on Linux. Eclipse substitutes space for {upload.setup_command}, and default is then applied by openocd, but Arduino IDE 1.8.7 let it unsubstituted in the command line. the upload.setup_command should be specified in board.txt

wodu:
I have already added it in the commit ab2fb3ca39ed183135310f56799a0f3b28b940b0

PR #333, opened 2019-01-17
boards: hackaBLE fix pin mapping

Pathak94:
There are two version of hackaBLE with different pins exposed.
This change makes nRF52832 pins map directly to Arduino pins
sequentially.

Signed-off-by: vaishali <vaishali@electronut.in>

### DIFF
--- a/variants/hackaBLE/variant.cpp
+++ b/variants/hackaBLE/variant.cpp
@@ -19,60 +19,38 @@
 
 #include "variant.h"
 
-/*
-                                Top view
-            RX - D0 | P0.14                 P0.27 | D7  - SDA
-            Tx - D1 | P0.13                 P0.26 | D8  - SCL
-                 D2 | P0.12                 P0.25 | D9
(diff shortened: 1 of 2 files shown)


Pathak94:
@dlabun, I thought about doing the same initially, but that will make things redundant. This change will make both the hackaBLE boards compatible with the same variant. We mapped each pin of Arduino controller to nordic's pin sequentially. This is the same as what Sparkfun breakout board has done for example, in their own fork of this.

Both boards have the pin numbers on the silkscreen as 'P0.21' etc. so this is easy to reference, and I have used only the pins that are common on both for Serial/SPI/I2C/LED_BUILTIN/BUTTON etc.

dlabun:
I am highly concerned about breaking any existing projects with this change. If the boards are broken out differently then different variants is appropriate to maintain past functionality.

Pathak94:
Okay, I will create a new variant with this pin mapping. Thanks @dlabun

PR #335, opened 2019-01-20
Update compatibility mode for different F_CPU values

nsamala:
Title says it all. Let me know what you think! This useful for boards like the nRF52832 where the boards are set to run at 64 MHz

### DIFF
--- a/libraries/SPI/SPI.h
+++ b/libraries/SPI/SPI.h
@@ -145,14 +145,13 @@ extern SPIClass SPI1;
 
 // For compatibility with sketches designed for AVR @ 16 MHz
 // New programs should use SPI.beginTransaction to set the SPI clock
-#if F_CPU == 16000000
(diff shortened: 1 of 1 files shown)


dlabun:
That's old compatibility code that nobody should be using now a days.... Do you have a use case for changing it?

nsamala:
@dbalun literally just for some test cases to pass for merging https://github.com/nsamala/platform-nordicnrf52/tree/feather_nrf52832 into PlatformIO

dlabun:
I'm going to leave this one for Sandeep to make the final call on, I am not of the opinion that we should be making changes to accommodate PlatformIO.

nsamala:
@dlabun the existing compatibility mode won't change. This just ensures that you don't hardcode the F_CPU into build files (which is what the maintainers of the PIO library had done) for this code to work. This is applicable outside of PlatformIO as well.

PR #336, opened 2019-01-21
fix for "undefined reference to `vtable for HardwareSerial'"

EndlessDelirium:
Not all functions of HardwareSerial are declared as pure virtual. This leads to "undefined reference to `vtable for HardwareSerial'" errors when used via Segger Studio.

### DIFF
--- a/cores/nRF5/HardwareSerial.h
+++ b/cores/nRF5/HardwareSerial.h
@@ -67,9 +67,9 @@
 class HardwareSerial : public Stream
 {
   public:
-    virtual void begin(unsigned long);
-    virtual void begin(unsigned long baudrate, uint16_t config);
-    virtual void end();
+    virtual void begin(unsigned long) = 0;
+    virtual void begin(unsigned long baudrate, uint16_t config) = 0;
+    virtual void end() = 0;
     virtual int available(void) = 0;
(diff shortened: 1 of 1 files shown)


dlabun:
Can you clarify what you mean by used via Segger Studio? Are building code inside their IDE or just debugging? The code builds without issue in Arduino IDE.

EndlessDelirium:
Yes, I'm building your project with Segger Studio.
Your project is awesome as it allows me to tap into the large Arduino ecosystem. The downside however is that I'm supposed to use the crappy Arduino IDE, reconfigure my J-Link and give up its debugging capabilities. Why should I do that when I can simply cherry pick all the good things?
This is the only code change needed in order to integrate this Arduino core into a Segger Studio project (apart from project configuration). It's rather unclear to me why some functions of HardwareSerial are pure virtual while others aren't. This change makes things more consistent and shouldn't lead to issues when building via Arduino IDE. It would however save other people who would like to integrate your Arduino core into a Segger Studio project some time.

dlabun:
I think this issue really originates from the ArduinoCore-Samd repository as it looks like Sandeep copied over HardwareSerial.h without any modifications. I have no problem merging in this fix, but you may want to also submit a patch to Arduino since it originated from them.

PR #340, opened 2019-02-07
ID115 fitness bracelet board support

EndlessDelirium:
The ID115 is a NRF51 based fitness tracker that is sometimes sold for as little as 8 USD. The combination of low cost, compact size and a rechargeable battery makes it rather compelling for little hacking projects.

Credits for this pull request go to @rbaron. What's submitted here is basically a copy from his repository. I only removed the LF Crystal option in boards.txt because this hardware doesn't have such a LF crystal.

More information about the hardware including an adapted driver for the OLED can be found on his blog: https://rbaron.net/blog/2018/05/27/Hacking-a-cheap-fitness-tracker-bracelet.html

For people interested in ordering this hardware it's worth pointing out that there are slightly different hardware revisions (the one I've got is nRF51802 based and there is a minor difference in the OLED memory layout too). Also some shops sell an entirely different hardware with a chinese BLE SoC as ID115, so compare the product pictures carefully before ordering and be prepared that things might not just work out of the box.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1024,3 +1024,50 @@ SeeedArchLink.menu.softdevice.s130.upload.maximum_size=151552
 SeeedArchLink.menu.softdevice.s130.build.extra_flags=-DNRF51 -DS130 -DNRF51_S130
 SeeedArchLink.menu.softdevice.s130.build.ldscript=armgcc_s130_nrf51822_xxaa.ld
 
+ID115.name=ID115
+
+ID115.upload.tool=sandeepmistry:openocd
+ID115.upload.target=nrf51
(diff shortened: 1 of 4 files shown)


dlabun:
I'm going to close this PR purely because we don't include any of the smart watches in this project. Mark Cooke would probably appreciate this PR on his version of this project as it focuses sole on nRF based smart watches:

https://github.com/micooke/arduino-nRF5-smartwatches

EndlessDelirium:
No problem. Maybe it would be worth pointing that out in the "supported boards" chapter of the readme?

dlabun:
Good idea, I'll add that on the next release.

PR #127, opened 2017-03-27
Update README.md

kriswiner:

PR #349, opened 2019-03-19
Add _BV

asukiaaa:
I want to use a library but it fails build on arduino-nRF5 because of missing `_BV`.
Could add it?

Thank you.

References:
[error: '_BV' was not declared in this scope](https://github.com/esp8266/Arduino/issues/125)
[Merge pull request #127 from Links2004/esp8266](https://github.com/esp8266/Arduino/commit/f1e6a72d8cb628cd8fa82d6ee11e97acef469b17)


dlabun:
_BV would indicate that’s a library written for an AVR based Arduino. Since _BV isn’t included in the Arduino SAMD core either I am going to reject this and recommend to fix the library to use bit instead.

asukiaaa:
I see.
Thank you for the feedback.

PR #351, opened 2019-03-26
Added initial nrf52840 support

mysensors-rus:
A lot of people want to build their IoT projects on nrf52840. This is a good CPU, you need to include it in support.


dlabun:
The nRF52840 is not compatible with the SDK or Soft devices included in this project so it's not possible to include the 840 at this time. I am also concerned with the quality of this PR, you removed the Generic nRF52832 without explaining why that change was made.

mysensors-rus:
I did not clean nrf52832. Just Generic nrf52832 already was, and I had to add Generic nrf52840. So I decided that it would be better to combine them and select the type of 
processor in the menu.
pr_scr01
SDK and Softdevices are needed by no means everyone. Many need simple support for nrf52840 as I suggested. In the proposed embodiment, port0, radio module, uarte0, timers are working. Why do you think that this is not enough for the initial support of nrf52840?

dlabun:
By changing / eliminating the generic nRF52832 variant you just broke any existing projects using that variant... That's a big issue since there was nothing wrong with the nRF52832 variant and there's nothing telling users about the change. There is no way the radio module is working as the nRF52840 doesn't support the S132 v2.0.1 SD used in this project. Every board in this project offers the exact same level of functionality... I am not thrilled by the idea of including partial support for the 840 and then having to explain why it doesn't work like every other board.

There's also other less serious issues such as the Readme and Travis CI script not being updated. If Travis isn't passing I basically stop reading the PR.

mysensors-rus:
That is, because of the identical structure of all the boards, you are denying the possibility of using nrf52840 to a huge number of people? I understand correctly that it makes no sense to convert PR into a separate Generic_nrf52840 board with an SD 140, will you still not accept it? Will you understand that not everyone needs SD?

dlabun:
No I’m saying don’t break the Generic nRF52832 variant while trying to add a new variant. If you want to add the nRF52840 you need to do it correctly. I will not merge a half baked PR that’s going to cause more trouble than good.

mysensors-rus:
Good. Old board back, added a new one.

dlabun:
You only updated boards.txt but you still need to create the rest of the new variant (ie, new folder, board specific variant.cpp and variant.h). You also need to update the ReadMe and TravisCI. Please resubmit when you have a complete PR for a new variant. If you don't understand what it takes to make a new variant there's plenty of other accepted PRs for new variants to look at.

PR #356, opened 2019-04-14
Fix #355 - Wrong SPI pins defined

### LINKED ISSUE #355: Wrong pin defined for hackaBLE SPI SS?
https://github.com/sandeepmistry/arduino-nRF5/blob/acd234f3ec70cdbca9c725ef15ec2a63da8e3801/variants/hackaBLE/variant.h#L85-L87

Perhaps SS should not be the same as MOSI?

henrygab:
Fix #355


dlabun:
Thanks for the fix

PR #386, opened 2019-11-01
Define a generic API for Timer2

pmalasp:
Timer 2 is usally available for application, but is constrained by the use of the common  ISR callback function TIMER2_IRQHandler which cannot be redefined across different libraries.

with this approach you can register up to 4 different callback functions for the 4 different CC registers, even if constrained by a single prescaler value.

Tested developing a simple app which uses modified  Adafruit_Microbit and IRremote libraries.


sandeepmistry:
Hi @pmalasp,

Thank you for this pull request. I would prefer that new functionality that is not part of the original Arduino API not be included in the project.

Instead I think it would be great if you could make a new Arduino for these new functionalities.

PR #432, opened 2020-10-28
BBC microbit v2

jhmaloney:
This pull request adds support for the BBC micro:bit v2, which is based on the nrRF2833. In addition to adding a new variant and a new entry in boards.txt, it also extends wiring-digital.c an the Wire library to handle pin numbers >31 (i.e. pins in the second GPIO port) since the micro:bit v2 uses some of those pins. It also adds a linker script for the nrRF2833.

The softdevice menu currently only provides "none" as an option because I'm not sure how to configure the soft device options for the  nrRF2833. I'm hoping someone else will fill in those options. Or I can give it a shot as a separate PR.

A final note is that the board definition file sets the CPU frequency to 64 MHz (BBCmicrobitV2.build.f_cpu=64000000). That seems correct even though other nRF52 boards that run at 64 MHz have that entry set to 16 MHz.

These changes have been tested with both the Arduino IDE 1.8.13 and with PlatformIO.


jhmaloney:
Fix: swapped SDA/SCL pins for external I2C

cores/nRF5/wiring_digital.c
```diff
@@ -18,27 +18,39 @@
...
 
 #include "nrf.h"
-
 #include "Arduino.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define GPIO_PORT0 ((NRF_GPIO_Type*) 0x50000000UL)
```
sandeepmistry, line 27: I have an idea that might simplify this, using the `digitalPinToPort` and `digitalPinToBitMask` macros in `/cores/nRF5/Arduino.h`. I'll make them and try to push to your branch so you can try them out.

libraries/Wire/Wire_nRF52.cpp
```diff
@@ -394,16 +405,29 @@ void TwoWire::onService(void)
   }
 }
 
-TwoWire Wire(NRF_TWIM1, NRF_TWIS1, SPIM1_SPIS1_TWIM1_TWIS1_SPI1_TWI1_IRQn, PIN_WIRE_SDA, PIN_WIRE_SCL);
+TwoWire Wire(NRF_TWIM0, NRF_TWIS0, SPIM0_SPIS0_TWIM0_TWIS0_SPI0_TWI0_IRQn, PIN_WIRE_SDA, PIN_WIRE_SCL);
```
sandeepmistry, line 408: This will be a breaking change, should I suggest that remain `NRF_TWIM1` etc here, so that `Wire` and `SPI` libs can be used at the same time. For `Wire1` using `NRF_TWIM0` etc. is fine.

package.json
```diff
@@ -0,0 +1,16 @@
+{
```
sandeepmistry, line 1: Please remove this file, it is not needed for Arduino cores.

variants/BBCmicrobitV2/variant.cpp
```diff
@@ -0,0 +1,58 @@
+/* Pins for micro:bit V2 */
+
+#include "variant.h"
+
+const uint32_t g_ADigitalPinMap[] = {
```
sandeepmistry, line 5: Do you have a link to the pin out diagram of the board? (Just want to double check this).

sandeepmistry (commented):
@jhmaloney thanks for submitting this! It's a great start and much more minimal a change than I was expecting!

I have a few inline comments to discuss and will try to push a commit to this branch to try to resolve one of them.

Again excellent work!

libraries/Wire/Wire_nRF52.cpp
jhmaloney: The micro:bit v2, unlike the original micro:bit, uses two separate i2c busses: one for the internal i2c devices and one for external devices, so you need both since the user might want to use the internal accelerometer and an external i2c device at the same time. However, I took a quick look at the nRF52833 manual and it appears to have three dedicated SPI peripherals and the nRF51 has two. So it may be possible to use the SPI peripheral for the SPI library on nRF52 boards. I'm not sure how bit a job that would be to implement.

package.json
jhmaloney: Will do.

libraries/Wire/Wire_nRF52.cpp
sandeepmistry: Ok, I think you just need to revert the changes you made to `Wire` to maintain backwards compatibility and have `Wire1` as:

```
TwoWire Wire1(NRF_TWIM0, NRF_TWIS0, SPIM0_SPIS0_TWIM0_TWIS0_SPI0_TWI0_IRQn, PIN_WIRE_SDA, PIN_WIRE_SCL);
```

cores/nRF5/wiring_digital.c
sandeepmistry: @jhmaloney I've pushed my changes in https://github.com/sandeepmistry/arduino-nRF5/pull/432/commits/652e38ab92eab4fcbfda547f08610dacec73b9fb and https://github.com/sandeepmistry/arduino-nRF5/pull/432/commits/3d82ada68832ed7b074f8d52501d54bc33409ae2 please try to pull and test if `pinMode`, 'digitalWrite`, `digitalRead` and `Wire` still work on your v2 board. I've tested on my v1 and v1.5 and things seem ok.

variants/BBCmicrobitV2/variant.cpp
jhmaloney: https://tech.microbit.org/hardware/schematic/

I don't believe there are any constraints on the ordering of the internal pins (i.e. user pin numbers > 20) so I've ordered them in a way that seems to make sense and be convenient for users. For example, the LED matrix pins for rows 1-5 are 21-25, so the second digit matches the row number.

I'd like the logo pin, speaker pin, and microphone power pin to follow the ROW pins immediately so MicroBlocks can use a simple comparison to check the range of user manipulable pins. (We don't want people to accidentally access the internal i2c or RX/TX pins since doing that would interfere with other functions. Also, they shouldn't use the microphone as a digital output since that could damage the microphone chip. They should only read the microphone as an analog input (A6). So all of those special pins should be located above the last manipulable pin (pin 28, RUN_MIC in the current pin map).

I also have a preliminary schematic but that was shared with me under NDA. If you want it, I'm sure Mark Williams or Jonny would send it to you.

cores/nRF5/wiring_digital.c
jhmaloney: Great! I see you've already made that change so I'll pull it and give it a try.

libraries/Wire/Wire_nRF52.cpp
jhmaloney: My understanding of the low-level details is limited, especially the interaction between SPI and TWI... 

Are you saying to use NRF_TWIM0, NRF_TWIS0, and SPIM0_SPIS0_TWIM0_TWIS0_SPI0_TWI0_IRQn for Wire1 and  NRF_TWIM1, NRF_TWIS1, and SPIM1_SPIS1_TWIM1_TWIS1_SPI1_TWI1_IRQn for Wire0? Is the issue that the interrupt vector is shared between TWI and SPI?

It might be best for you to make this change. The solution should allow use of both Wire and Wire1 at the same time (or, more accurately, interleaved). Ideally, it would also allow use of SPI as well.

cores/nRF5/wiring_digital.c
jhmaloney: I found a fixed two small bugs involving pins on port1. Also fixed a small bug in Wire_nRF52.cpp. With these fixes digitalWrite and the internal I2C bus work. I haven't yet tested digitalRead and the external I2C bus.

I've also made two small changes to the pin map for the Calliope mini: (1) added the ability to use connector pins 16 and 17 as analog inputs A6 and A7; and (2) swapped the positions of the microphone pin and an unassigned pin in the digital pin map. Since the microphone read as A0 and should never be used as a digital pin, change (2) should not impact any existing code. This change makes it easier to prevent digital access to the microphone pin (i.e. using a simple range check) since using that pin as a digital output could potentially damage the microphone amplifier (e.g. setting the pin to 0 would connect the output of the amplifier to ground).

jhmaloney: Have now tested digitalRead on pins from both ports and tested the external i2c bus. In the process, I found and fixed a bug in i2c handling of port1 pins).

I also found and fixed a long-standing bug in the pulseIn() function. (It should have been mapping the user level pin number to the internal pin number.)

Things are looking good!

jhmaloney: I fixed what I believe was a bug in pulseIn(). It should take a user pin number (e.g. the numbers used in digitalRead()) but some old (and incorrect) code to do that mapping was commented out. The computation of pulse microseconds from loops was also off by about 30%. I adjusted the constant based on tests against a range of pulse widths that were checked with an oscilloscope. Since pulse width measurements are quantized to an integral number of loops and subject to jitter due to interrupts, this function isn't expected to have microsecond precision, but I think the new calculation is much closer than the old one. (One use-case is decoding the signal from a DHT humidity/temperature sensor, which requires distinguishing between pulse widths of 28 and 70 usecs.)

luqmaanb:
Hi

Great work on this pull request. I've managed to use I2C successfully with SSD1306 but I can't get SPI to work? I'm using the [IM01](https://xinabox.cc/products/im01?_pos=1&_sid=4f95806d3&_ss=r) bridge from XinaBox that allows for SD card operations. Its working perfectly on microbit 1 but freezes at SD card initialization on microbit 2. The example I'm using the one from Arduino

```c
/*
  SD card test

  This example shows how use the utility libraries on which the'
  SD library is based in order to get info about your SD card.
  Very useful for testing a card when you're not sure whether its working or not.

  The circuit:
    SD card attached to SPI bus as follows:
 ** MOSI - pin 11 on Arduino Uno/Duemilanove/Diecimila
 ** MISO - pin 12 on Arduino Uno/Duemilanove/Diecimila
 ** CLK - pin 13 on Arduino Uno/Duemilanove/Diecimila
 ** CS - depends on your SD card shield or module.
 		Pin 4 used here for consistency with other Arduino examples


  created  28 Mar 2011
  by Limor Fried
  modified 9 Apr 2012
  by Tom Igoe
*/
// include the SD library:
#include <SPI.h>
#include <SD.h>

//#include <xOD01.h>

//xOD01 OD01;

// set up variables using the SD utility library functions:
Sd2Card card;
SdVolume volume;
SdFile root;

// change this to match your SD shield or module;
// Arduino Ethernet shield: pin 4
// Adafruit SD shields and modules: pin 10
// Sparkfun SD shield: pin 8
// MKRZero SD: SDCARD_SS_PIN
const int chipSelect = 16;

void setup() {
  // Open serial communications and wait for port to open:
  Serial.begin(115200);
  while (!Serial) {
    ; // wait for serial port to connect. Needed for native USB port only
  }
//Wire.begin();
//OD01.begin();
//OD01.clear();


  Serial.print("\nInitializing SD card...");

  // we'll use the initialization code from the utility libraries
  // since we're just testing if the card is working!
  if (!card.init(SPI_HALF_SPEED, chipSelect)) {
    Serial.println("initialization failed. Things to check:");
    Serial.println("* is a card inserted?");
    Serial.println("* is your wiring correct?");
    Serial.println("* did you change the chipSelect pin to match your shield or module?");
    while (1);
  } else {
    Serial.println("Wiring is correct and a card is present.");
  }

  // print the type of card
  Serial.println();
  Serial.print("Card type:         ");
  switch (card.type()) {
    case SD_CARD_TYPE_SD1:
      Serial.println("SD1");
      break;
    case SD_CARD_TYPE_SD2:
      Serial.println("SD2");
      break;
    case SD_CARD_TYPE_SDHC:
      Serial.println("SDHC");
      break;
    default:
      Serial.println("Unknown");
  }

  // Now we will try to open the 'volume'/'partition' - it should be FAT16 or FAT32
  if (!volume.init(card)) {
    Serial.println("Could not find FAT16/FAT32 partition.\nMake sure you've formatted the card");
    while (1);
  }

  Serial.print("Clusters:          ");
  Serial.println(volume.clusterCount());
  Serial.print("Blocks x Cluster:  ");
  Serial.println(volume.blocksPerCluster());

  Serial.print("Total Blocks:      ");
  Serial.println(volume.blocksPerCluster() * volume.clusterCount());
  Serial.println();

  // print the type and size of the first FAT-type volume
  uint32_t volumesize;
  Serial.print("V type is:    FAT");
  Serial.println(volume.fatType(), DEC);

  volumesize = volume.blocksPerCluster();    // clusters are collections of blocks
  volumesize *= volume.clusterCount();       // we'll have a lot of clusters
  volumesize /= 2;                           // SD card blocks are always 512 bytes (2 blocks are 1KB)
  Serial.print("V size (Kb):  ");
  Serial.println(volumesize);
  Serial.print("V size (Mb):  ");
  volumesize /= 1024;
  Serial.println(volumesize);
  Serial.print("V size (Gb):  ");
  Serial.println((float)volumesize / 1024.0);

  Serial.println("\nFiles found on the card (name, date and size in bytes): ");
  root.openRoot(volume);

  // list all files in the card with date and size
  root.ls(LS_R | LS_DATE | LS_SIZE);
}

void loop(void) {
}
```

Have you successfully tested the SPI? I'd like to rule out an SPI issue.

jhmaloney:
I haven't tested SPI and, in fact, Sandeep mentioned that there could be an issue in the declaration of Wire1. You could tried changing that to:
`TwoWire Wire1(NRF_TWIM0, NRF_TWIS0, SPIM0_SPIS0_TWIM0_TWIS0_SPI0_TWI0_IRQn, PIN_WIRE1_SDA, PIN_WIRE1_SCL);`
Note that that might make it impossible to use Wire1 and Wire at the same time. They are both needed because the V2 uses separate internal and external I2C buses.

luqmaanb:
> I haven't tested SPI and, in fact, Sandeep mentioned that there could be an issue in the declaration of Wire1. You could tried changing that to:
> 
> `TwoWire Wire1(NRF_TWIM0, NRF_TWIS0, SPIM0_SPIS0_TWIM0_TWIS0_SPI0_TWI0_IRQn, PIN_WIRE1_SDA, PIN_WIRE1_SCL);`
> 
> Note that this change may make it impossible to use Wire1 and Wire at the same time. Both those are needed because the V2 uses separate internal and external I2C buses.
> 
> A better solution might be to change the SPI library to use SPI2 when running on an nRF52. It looks like that that avoids the conflict with both TWI0 and TWI1. The nRF51 might be limited to a single Wire instance (i.e. Wire1 could be #ifdef-ed out except on nRF52's.) @sandeep, does that make sense?

Thanks. Changing that line of code made the SPI work but as you said, the I2C stopped working. I did, however, have to add these few lines in the `variant.h` file:

```c
static const uint8_t SS   = 16 ;
static const uint8_t MOSI = PIN_SPI_MOSI ;
static const uint8_t MISO = PIN_SPI_MISO ;
static const uint8_t SCK  = PIN_SPI_SCK ;
```

I'm looking into the spi/i2c conflict. Will post if I get anything working.

variants/BBCmicrobitV2/variant.cpp
```diff
@@ -0,0 +1,58 @@
...
+  5,  // A6, MIC_IN
+
+  // 30 - 31
+  16, // I2C_INT_SDA
+  8,  // I2C_INT_SCL
+
+  // 32 - 34
+  25, // COMBINED_SENSOR_INT
+  6,  // RX
+  40, // TX, P1.08
```
sandeepmistry, line 57: Should the comments be swapped here?

Below in `variants/BBCmicrobitV2/variant.h` you have:
```
#define PIN_SERIAL_RX		(34)
#define PIN_SERIAL_TX		(33)
```

jhmaloney: Yes, I think so. The comments come from the pin listing in micro:bit technical documentation:

  https://tech.microbit.org/hardware/schematic/

But I think the pin listing is incorrect. The RX pin on the nRF52 should connected to the TX pin of the KL27 and vice versa. The pin listing has TX to TX and RX to RX. I will email the micro:bit folks to double-check.

sandeepmistry:
@luqmaanb @jhmaloney could you please try out PR https://github.com/sandeepmistry/arduino-nRF5/pull/439, I've updated the nRF MDK so it's easier to add support for the micro:bit v2 (based on @jhmaloney's initial changes).

sandeepmistry:
Closing in favour of https://github.com/sandeepmistry/arduino-nRF5/pull/439.

PR #407, opened 2020-03-07
Fix #312: softdevice download links should be stable now.

honnet:
Hi all!
As you can see in the commit, the new URLs look more "durable", so this commit should fix the download problem for a while.

Note: to avoid changing any behavior, this commit only allows downloading the exact same versions of the SD, but there are much more recent ones:
https://www.nordicsemi.com/Software-and-tools/Software/S132/Download#infotabs
https://www.nordicsemi.com/Software-and-tools/Software/S130/Download#infotabs

=> If any of you is confident about the retro-compatibility of the new SD, I'd be happy to improve this pull request.

Thanks again for this great project!
Cedric ;)

PR #433, opened 2020-11-03
Update the urls to the Nordic Softdevice hex files

KevinJohnMulligan:
I have found the correct links for the zip files. The older links are now dead and give a 404 error.

I have checked the zip files and despite having a very specific name they still contain a generically named *.hex* file that is consistent with the previous names 

For example: `s110nrf51800.zip` contains `s110_nrf51_8.0.0_softdevice.hex`

### DIFF
--- a/softdevices.txt
+++ b/softdevices.txt
@@ -1,10 +1,10 @@
 names=s110,s130,s132
 
-s110.url=http://www.nordicsemi.com/eng/content/download/80234/1351257/file/s110_nrf51_8.0.0.zip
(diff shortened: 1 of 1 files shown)


sandeepmistry:
Hi @KevinJohnMulligan,

Thanks for the PR, I believe it's the same changes as https://github.com/sandeepmistry/arduino-nRF5/pull/407 which I just merged. I'll close it for now, but let me know if I'm missed something.

KevinJohnMulligan:
Hi @sandeepmistry,

Thank you for your response.

Yes the changes appear to be the same.

PR #442, opened 2020-12-19
Add support for Generic NRF52805

nikolac:
I made a couple of questionable decisions:
1. I'm not 100% sure how to determine maximum upload size. I set 192000.
2.  This chip does not support PWM, so in addition to making all of the internal PWM references conditional based PWM_PRESENT, I also wrapped the contents of analogWrite with the condition. Would it be more appropriate to make the analogWrite definition conditional? Probably better to throw compiler errors if somebody tried to use it?

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -21,6 +21,38 @@ menu.version=Version
 menu.lfclk=Low Frequency Clock
 menu.board_variant=Board Variant
 
+# nRF52805 variants
+###################
+
+Generic_nRF52805.name=Generic nRF52805
+
+Generic_nRF52805.upload.tool=sandeepmistry:openocd
+Generic_nRF52805.upload.target=nrf52
+Generic_nRF52805.upload.maximum_size=192000
+
+Generic_nRF52805.bootloader.tool=sandeepmistry:openocd
+
+Generic_nRF52805.build.mcu=cortex-m4
+Generic_nRF52805.build.f_cpu=64000000
+Generic_nRF52805.build.board=GENERIC
+Generic_nRF52805.build.core=nRF5
+Generic_nRF52805.build.variant=Generic
+Generic_nRF52805.build.variant_system_lib=
+Generic_nRF52805.build.extra_flags=-DNRF52805_XXAA
+Generic_nRF52805.build.float_flags=
+Generic_nRF52805.build.ldscript=nrf52805_xxaa.ld
+
+Generic_nRF52805.menu.softdevice.none=None
+Generic_nRF52805.menu.softdevice.none.softdevice=none
+Generic_nRF52805.menu.softdevice.none.softdeviceversion=
+
+Generic_nRF52805.menu.lfclk.lfxo=Crystal Oscillator
+Generic_nRF52805.menu.lfclk.lfxo.build.lfclk_flags=-DUSE_LFXO
+Generic_nRF52805.menu.lfclk.lfrc=RC Oscillator
+Generic_nRF52805.menu.lfclk.lfrc.build.lfclk_flags=-DUSE_LFRC
+Generic_nRF52805.menu.lfclk.lfsynt=Synthesized
+Generic_nRF52805.menu.lfclk.lfsynt.build.lfclk_flags=-DUSE_LFSYNT
+
 # nRF52833 variants
 ###################
 
--- a/boards.txt
+++ b/boards.txt
@@ -28,7 +28,7 @@ Generic_nRF52805.name=Generic nRF52805
 
 Generic_nRF52805.upload.tool=sandeepmistry:openocd
 Generic_nRF52805.upload.target=nrf52
-Generic_nRF52805.upload.maximum_size=192000
+Generic_nRF52805.upload.maximum_size=196608
 
 Generic_nRF52805.bootloader.tool=sandeepmistry:openocd
 
--- a/cores/nRF5/Uart.cpp
+++ b/cores/nRF5/Uart.cpp
@@ -64,12 +64,22 @@ void Uart::begin(unsigned long baudrate)
 
 void Uart::begin(unsigned long baudrate, uint16_t /*config*/)
 {
+#ifdef NRF52805_XXAA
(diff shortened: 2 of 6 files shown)


libraries/Wire/Wire_nRF52.cpp
```diff
@@ -401,17 +401,34 @@ void TwoWire::onService(void)
   }
 }
 
-TwoWire Wire(NRF_TWIM1, NRF_TWIS1, SPIM1_SPIS1_TWIM1_TWIS1_SPI1_TWI1_IRQn, PIN_WIRE_SDA, PIN_WIRE_SCL);
+#if defined(NRF52805_XXAA)
```
sandeepmistry, line 404: Could also consider `#if defined(NRF_TWIM1` here as well.

sandeepmistry (commented):
@nikolac great work!

I do not have an nRF52085 based device board to try these changes out with. However, I made a few inline comments.

cores/nRF5/Uart.cpp
```diff
@@ -64,12 +64,22 @@ void Uart::begin(unsigned long baudrate)
 
 void Uart::begin(unsigned long baudrate, uint16_t /*config*/)
 {
+#ifdef NRF52805_XXAA
```
sureshsitaula, line 67: I was working on nrf52810 and I started with adding NRF528**10**_XXAA to this ifdef (and that works) but realized that complicated things when you need to support multiple chips ...

My [solution](https://github.com/sureshsitaula/arduino-nRF5/commit/0bd1c2aa31e5a8cb0c5dc3e07f07f9b9a1c186f8) was to edit the compat header (cores/nRF5/SDK/components/device/nrf52_to_nrf52810.h) instead and I didn't need to modify this file.
(and this matches the existing implementation for some other chips, see [nrf52840](https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/SDK/components/device/nrf52_to_nrf52840.h) for example).

Should we update the compat header instead of this file?

libraries/SPI/SPI.cpp
```diff
@@ -256,7 +256,7 @@ void SPIClass::detachInterrupt() {
 }
 
 #if SPI_INTERFACES_COUNT > 0
-#if defined(NRF52_SERIES)
+#if defined(NRF52_SERIES) && !defined(NRF52805_XXAA)
```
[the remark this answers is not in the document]
sureshsitaula: +1 That would make it work on nrf52810 as well.

cores/nRF5/Uart.cpp
nikolac: Should we be touching code in the SDK? Does that introduce potential upgrade complexity?

libraries/Wire/Wire_nRF52.cpp
nikolac: I believe the NRF52805 has completely different handler names (TWIM0_TWIS0_TWI0_IRQn vs SPIM0_SPIS0_TWIM0_TWIS0_SPI0_TWI0_IRQn) which was part of the reason for using the board specific logic.

cores/nRF5/Uart.cpp
sureshsitaula: Good question for @sandeepmistry :)
It would be nice to have this consistency between compat .h files, like [nrf52_to_nrf52840.h](https://github.com/sandeepmistry/arduino-nRF5/blob/master/cores/nRF5/SDK/components/device/nrf52_to_nrf52840.h) referenced above.

sandeepmistry: > Should we be touching code in the SDK? Does that introduce potential upgrade complexity?

It would be better to avoid doing this, as it would be one more thing to manage when updating the SDK.

That said if it's the best approach wrapping changes in `#ifdef ARDUINO` would be best.

libraries/Wire/Wire_nRF52.cpp
sandeepmistry: Got it, sorry for not checking this out before!

libraries/SPI/SPI.cpp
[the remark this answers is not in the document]
nikolac: Done here https://github.com/sandeepmistry/arduino-nRF5/pull/442/commits/110910239fad390fcc3d9abfcfde84f11df1d041

boards.txt
```diff
@@ -21,6 +21,38 @@ menu.version=Version
...
 menu.board_variant=Board Variant
 
+# nRF52805 variants
+###################
+
+Generic_nRF52805.name=Generic nRF52805
+
+Generic_nRF52805.upload.tool=sandeepmistry:openocd
+Generic_nRF52805.upload.target=nrf52
+Generic_nRF52805.upload.maximum_size=192000
```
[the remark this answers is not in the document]
nikolac: Fixed here https://github.com/sandeepmistry/arduino-nRF5/pull/442/commits/a60f6db675d687bf964930a0505a94a1a9ad17d6

cores/nRF5/Uart.cpp
nikolac: Am I good to leave this as-is for now? There is already series and in some special cases chip specific logic in the .cpp library files. I would prefer to leave this condition where it is until we decide on the best way/place to extend SDK definitions without modifying the SDK.

However, I'll do whatever you feel is best. There is increased interested to get this merged in, so whatever I can do to help expedite.

nikolac:
Just wanted to follow up on this. I've been using this branch for ~2 years.

tobigun:
What is the status of this pull-request?
I have solved the conflicts in my branch and also implemented analogWrite() the same way it was done in nRF51:
https://github.com/tobigun/arduino-nRF5/tree/support-nrf52805

spike314:
Thanks for this fork nikolac.  I know this PR is a bit stale, but it was very helpful.  I was frustrated using the NRF52810 because I could not get the I2C to work.  Turns out if follows the same pattern as the NRF52805, so I was able to use your updates.   Now have NRF52805 and NRF52810 nodes working.

PR #488, opened 2022-06-11
Fix Arduino.h `portOutputRegister()` and `portModeRegister()`

carlosperate:
These functions were using the `OUTSET` and `DIRSET` registers instead of `OUT` and `DIR`. The `SET` version of these registers only do something when a "1" is written to any of their bits, writting a "0" does nothing.

From the nRF52833 datasheet:
image
image

Found this issue when using some of the Adafruit libraries using a `BUSIO_USE_FAST_PINIO` macro that enables utilisation of these functions, and noticed that some of the pins were not being driven low, only high.
Using this patch locally fixed the issue.

### DIFF
--- a/cores/nRF5/Arduino.h
+++ b/cores/nRF5/Arduino.h
@@ -109,9 +109,9 @@ void loop( void ) ;
 #error "Unsupported GPIO_COUNT"
 #endif
 
-#define portOutputRegister(port)   ( &(port->OUTSET) )
+#define portOutputRegister(port)   ( &(port->OUT) )
 #define portInputRegister(port)    ( &(port->IN) )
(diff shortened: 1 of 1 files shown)


carlosperate:
For comparison I've also found these other nRF cores using the registers from this PR:
- https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/1.3.0/cores/nRF5/Arduino.h#L148-L151
- https://github.com/arduino/ArduinoCore-primo/blob/1.0.1/variants/arduino_primo/variant.h#L67-L69
- https://github.com/particle-iot/device-os/blob/v3.3.0/user/inc/Arduino.h#L108-L118

So that should give us some confidence as well.

PR #489, opened 2022-06-12
CI: Add GitHub Actions workflow to build blinky with all variants.

carlosperate:
As TravisCI is no longer a good option for open source projects and it is no longer running for this repository.

In this case I opted to use GitHub Actions as it is readily available and fairly popular. And this workflow uses the official Arduino Action to compile sketches.

Also added 3 or 4 variants that were not in the original TravisCI list, they compile fine, so they must have been accidentally left out.


.github/workflows/compile.yml
```diff
@@ -0,0 +1,76 @@
...
+          verbose: true
+          enable-deltas-report: true
+          platforms: |
+            # First we have to install this core so that the tooling is installed
+            - name: "sandeepmistry:nRF5"
+              source-url: "https://sandeepmistry.github.io/arduino-nRF5/package_nRF5_boards_index.json"
+              version: latest
+            # Now, this will overwrite the core source to use the local version
+            - name: "sandeepmistry:nRF5"
+              source-path: .
```
carlosperate, line 74: It's a bit funny having to first install a released version of the core, since we need to compile with the local code, so we have to then "reasign" the core to point to the local version.

It looks like this might be a limitation of the Arduino Action, since the official Arduino cores do the same:
- https://github.com/arduino/ArduinoCore-mbed/blob/3.1.1/.github/workflows/compile-examples.yml#L138-L143
- https://github.com/arduino/ArduinoCore-avr/blob/1.8.5/.github/workflows/compile-platform-examples.yml#L186-L191

sandeepmistry:
@carlosperate these changes look good!

Should the build badge in the read me also be updated?

carlosperate:
Oh yeah, good catch!
I've replaced the badge and renamed the workflow "Build" as otherwsie the previous old name appears in the badge.

sandeepmistry (approved):

PR #500, opened 2022-12-29
added ST-Link/V2-1 aka ST-Link On-Board as stlinkob

so2akt:
added ST-Link/V2-1 on Programmers menu, below ST-Link V2

### DIFF
--- a/programmers.txt
+++ b/programmers.txt
@@ -12,6 +12,13 @@ stlink.program.protocol=stlink-v2
 stlink.program.tool=openocd
 stlink.program.setup_command=transport select hla_swd; set WORKAREASIZE 0x4000;
 
+stlinkob.name=ST-Link V2-1 (ST-Link On-Board)
+stlinkob.communication=USB
+stlinkob.protocol=stlink-v2-1
+stlinkob.program.protocol=stlink-v2-1
+stlinkob.program.tool=openocd
+stlinkob.program.setup_command=transport select hla_swd; set WORKAREASIZE 0x4000;
+
 cmsisdap.name=CMSIS-DAP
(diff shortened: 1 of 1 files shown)


so2akt:
sorry for mis-closed!

carlosperate:
What boards have this st-link on board?
Doesn't the "normal" stlink profile work?

so2akt:
Thank you for your reply!
This ST-Link, ST-Link/V2-1 is on STM32 [Nucleo (except for NUCLEO-32)](https://www.st.com/en/ecosystems/stm32-nucleo.html) and [DISCOVERY](https://www.st.com/en/evaluation-tools/stm32-discovery-kits.html) series. I use [NUCLEO-L053R8](https://www.st.com/ja/evaluation-tools/nucleo-l053r8.html) with SWD connection.
In addition, the "normal" stlink profile does not work because ST-Link/V2-1 has different PID as follows:

| debugger | VID | PID | OpenOCD config file | note |
| --- | --- | --- | --- | --- |
| ST-Link/V2 | 0x0483 | 0x374**8** | `stlink-v2.cfg` | NORMAL stlink |
| ST-Link/V2-1 | 0x0483 | 0x374**B** | `stlink-v2-1.cfg` | on-board on NUCLEO and DISCOVERY |

This spec is mentioned [this stack overflow article](https://stackoverflow.com/questions/31638347/stm32f0-st-link-v2-openocd-0-9-0-open-failed), and [STM 32 Nucleo-64 boards User Manual](https://www.st.com/resource/en/user_manual/um1724-stm32-nucleo64-boards-mb1136-stmicroelectronics.pdf), section 6.2.

PR #532, opened 2024-10-18
Creates the NRF52840DK variant

needs-coffee:
previous target for the NRF52840_dk used the NRF52DK variant
The NRF52DK variant pinout is specific to the NRF52832

This is the correct pinout for the NRF52840DK with arduino style mapping


albertmarc8 (approved):
I had been looking for a while for the fix on this issue. I've created an issue at https://github.com/platformio/platform-nordicnrf52/issues/208. This pull request along with https://github.com/platformio/platform-nordicnrf52/pull/200 has fixed this issue for me.

PR #533, opened 2024-10-18
Add a generic variant for the NRF52840DK board

needs-coffee:
Adds a generic variant for the NRF52840DK board that accesses all pins using the NRF GPIO index.
Has the on board peripherals mapped to the correct pin functions.


DaveVdE:
Hi,

First of all, thank you for the hard work of making a PR to add support for this board.

I'm a newbie at making BSPs but, shouldn't we be adding something to boards.txt as well? I've had a look at how it looks in the Adafruit BSP for this board but it seems the chosen programmer is entirely different from the nRF52 DK so I'm at a loss.

